PR #1925, opened 2021-09-29
Update portal.md

ajesien27:
Clarifying a very confusing and incorrect paragraph. We can only block Red data with the Privacy Portal. There is nothing functionally different between Green and Yellow classifications and you can't block these. We also don't have any destination-specific PII controls (outside of what's possible with Destination Filters)


pwseg (approved):
Thanks for correcting this.

markzegarelli (approved):

markzegarelli (approved):

PR #1996, opened 2021-10-11
Update FB Pixel and Conversions API docs

kdaswani:
### Proposed changes
- Update FB Pixel docs to explain requirements for using external_id
- Update FB Conversions API docs to explain requirements for using external_id client-side

### Merge timing
- ASAP once approved

### Related issues
https://segment.atlassian.net/browse/STRATCONN-968


markzegarelli:
> also, I noticed the branch chosen to merge into is `develop` instead of `master`. is that expected? thanks!

Hey @kdaswani, yes this is expected. We have transitioned to having twice-weekly docs releases. We do this by queuing work in `develop` and then merging to `master` at set times throughout the week.

PR #2016, opened 2021-10-13
DOC-335: SQL Traits FAQ fix

pwseg:
### Proposed changes

Corrected FAQ regarding how often Segment queries data warehouses.

### Merge timing

- ASAP once approved

### DIFF
--- a/src/personas/sql-traits.md
+++ b/src/personas/sql-traits.md
@@ -207,7 +207,7 @@ The result set is capped at 10 million rows.
 
 ### How often does Segment query the customer's data warehouse?
 
-Segment queries the data warehouse every 12 hours by default, but can query up to hourly. [Contact us](https://segment.com/help/contact/) for customized schedules.
+For each SQL Trait you create, you can set a compute schedule to query the data warehouse up to once per hour. Your query may run at any given time during the hour you select.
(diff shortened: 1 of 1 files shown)


src/personas/sql-traits.md
```diff
@@ -207,7 +207,7 @@ The result set is capped at 10 million rows.
 
 ### How often does Segment query the customer's data warehouse?
 
-Segment queries the data warehouse every 12 hours by default, but can query up to hourly. [Contact us](https://segment.com/help/contact/) for customized schedules.
+For each SQL Trait you create, you can set a compute schedule to query the data warehouse up to once per hour. Your query may run at any given time during the hour you select.
```
stayseesong, line 210:
```suggestion
For each SQL Trait you create, you can set a compute schedule to query the data warehouse once per hour. Your query may run at any given time during the hour you select.
```

pwseg: Hmm.  In this case I think the "up to" matters, because it doesn't *have* to be once per hour (it could be done less frequently than that, like once per day, or every twelve hours).

stayseesong: got it

PR #2150, opened 2021-11-16
Create an IAM role using the AWS CLI

forstisabella:
### Proposed changes

Added section to the existing [AWS S3 with IAM Role Support Destination](https://segment.com/docs/connections/storage/catalog/aws-s3/) page that covers creating an IAM role using the AWS CLI.

### Merge timing
On or before Thursday, November 18th

PR #2178, opened 2021-11-23
Updated procedure for creating an IAM role in the AWS console [DOC-375]

forstisabella:
### Proposed changes

Made the procedure for creating an IAM role from the AWS console clearer

### Related issues (optional)
#2150 #2164

### DIFF
--- a/src/connections/storage/catalog/aws-s3/index.md
+++ b/src/connections/storage/catalog/aws-s3/index.md
@@ -45,23 +45,23 @@ To complete this section, you need access to your AWS dashboard.
 3. Attach the following policy to the IAM role created in step 2. Replace `<YOUR_BUCKET_NAME>` with the name of the S3 bucket you created in step 1.    
     ```json 
     {
-      "Version": "2012-10-17",
-      "Statement": [
-          {
-          "Sid": "PutObjectsInBucket",
-          "Effect": "Allow",
-          "Action": [
-              "s3:PutObject",
-              "s3:PutObjectAcl"
-          ],
-          "Resource": "arn:aws:s3:::<YOUR_BUCKET_NAME>/segment-logs/*"
-          }
+        "Version": "2012-10-17",
+        "Statement": [
+            {
+            "Sid": "PutObjectsInBucket",
+            "Effect": "Allow",
+            "Action": [
+                "s3:PutObject",
+                "s3:PutObjectAcl"
+            ],
+            "Resource": "arn:aws:s3:::<YOUR_BUCKET_NAME>/segment-logs/*"
+            }
       ]
     }   
     ```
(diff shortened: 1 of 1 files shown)


rchinn-segment (approved):

src/connections/storage/catalog/aws-s3/index.md
```diff
@@ -94,8 +76,29 @@ To complete this section, you need access to your AWS dashboard.
       ]
     }
     ```
-
 If you have server-side encryption enabled, see the [required configuration](#encryption).
+ 
+4. Update `Principal` in the role’s trust relationship document to `arn:aws:iam::595280932656:role/segment-s3-integration-production-access`. Replace `<YOUR_WORKSPACE_ID>` with your Segment Workspace ID.
```
maggieyu-segment, line 81: Customer would have entered the workspace ID as the external ID in step 2.3. Just wondering if `Replace <YOUR_WORKSPACE_ID>...` may be confusing to some customers? (Not a change request - just wondering)

maggieyu-segment (approved):

src/connections/storage/catalog/aws-s3/index.md
forstisabella: That's a good point! I phrased it that way to keep it consistent with the other steps in this procedure that involved users replacing the objects in angle brackets with their own values, but if you think it would make more sense to say something like "Add the External ID from the IAM role in step 2.3," I'd be happy to include that instead!

maggieyu-segment: What you have written is fine - no need to change! Thanks!

src/connections/storage/catalog/aws-s3/index.md
```diff
@@ -39,63 +39,66 @@ To complete this section, you need access to your AWS dashboard.
...
+      "Version": "2012-10-17",
+      "Statement": [
+          {
+          "Sid": "PutObjectsInBucket",
+          "Effect": "Allow",
+          "Action": [
+              "s3:PutObject",
+              "s3:PutObjectAcl"
+          ],
+          "Resource": "arn:aws:s3:::<YOUR_BUCKET_NAME>/segment-logs/*"
```
maggieyu-segment, line 57: Indentation for these lines

src/connections/storage/catalog/aws-s3/index.md
```diff
@@ -39,63 +39,66 @@ To complete this section, you need access to your AWS dashboard.
...
+          }
+      ]
+    }   
     ```
     If you're using KMS encryption on your S3 bucket, add the following policy to the IAM role:
     ```json
+      {
```
maggieyu-segment, line 64: I think this bracket isn't starting from position 1, i.e. this has been indented but it doesn't need to be.

PR #2222, opened 2021-11-30
Clarify wording on timestamps + Personas note

sanscontext:
### Proposed changes

Not sure how old this note is, but I had to read it several times before I understood it. Reordered a bit to clarify. :)

### Merge timing

- once approved ¯\_(ツ)_/¯

### DIFF
--- a/src/connections/spec/common.md
+++ b/src/connections/spec/common.md
@@ -305,7 +305,7 @@ Every API call has four timestamps, `originalTimestamp`, `timestamp`, `sentAt` a
 **All timestamps are [ISO-8601](http://en.wikipedia.org/wiki/ISO_8601) date strings.**
 
 > note ""
-> **NOTE:** ISO-8601 date strings with timezones included are required when using timestamps with [Personas](/docs/personas/). Sending custom traits without a timezone included in the timestamp will result in the value not being saved.
+> **NOTE:** You must use ISO-8601 date strings _including_ timezones when you use timestamps with [Personas](/docs/personas/). If you send custom traits without a timezone, Segment does not save the timestamp value.
 
 ### Timestamp Overview
 


src/connections/spec/common.md
```diff
@@ -305,7 +305,7 @@ Every API call has four timestamps, `originalTimestamp`, `timestamp`, `sentAt` a
 **All timestamps are [ISO-8601](http://en.wikipedia.org/wiki/ISO_8601) date strings.**
 
 > note ""
-> **NOTE:** ISO-8601 date strings with timezones included are required when using timestamps with [Personas](/docs/personas/). Sending custom traits without a timezone included in the timestamp will result in the value not being saved.
+> **NOTE:** You must use ISO-8601 date strings _including_ timezones when you use timestamps with [Personas](/docs/personas/). If you send custom traits without a timezone, Segment does not save the timestamp value.
```
stayseesong, line 308:
```suggestion
> **NOTE:** You must use ISO-8601 date strings that include timezones when you use timestamps with [Personas](/docs/personas/). If you send custom traits without a timezone, Segment doesn't save the timestamp value.
```

src/connections/spec/common.md
```diff
@@ -305,7 +305,7 @@ Every API call has four timestamps, `originalTimestamp`, `timestamp`, `sentAt` a
 **All timestamps are [ISO-8601](http://en.wikipedia.org/wiki/ISO_8601) date strings.**
```
stayseesong, line 305:
```suggestion
**All timestamps are [ISO-8601](http://en.wikipedia.org/wiki/ISO_8601){:target="_blank"} date strings.**
```

stayseesong (approved):

stayseesong:
thanks for the suggestion @sanscontext! Good to hear from you!

PR #2264, opened 2021-12-09
Friendbuy destination

markzegarelli:
- Friendbuy documentation initial version.
- Update for Adam's feedback.


src/connections/destinations/catalog/actions-friendbuy-web/index.md
```diff
@@ -0,0 +1,129 @@
...
+
+To configure a custom event:
+
+1. Click **New Mapping** and choose **Track Custom Event**.
+
+2. For the Event Name, choose the name of the Segment event you are tracking. For example, if you want to reward on app downloads that you are tracking with `analytics.track("Downloaded", { downloadId: "My App" }`, select `Downloaded` as your *Event Name*.
+
+3. When you configure your action fields, you can accept the default *Event Name* of `event` which will cause Friendbuy to receive track events with the same names as the Segment events, or if you want to rename the Friendbuy events you can replace the *Event Name* with a different value such as the static string `download`.
+
+4. By default the *Event Properties* is set from the path `properties`, which means that all the properties that you include in your Segment `analytics.track` call will be passed to Friendbuy. If you don't want to do that, you will have to segregate the fields to pass to Friendbuy into an object and then specify the path to that object similar to what is described above under [Custom Attributes](#custom-attributes) (except in this case the *only* attributes that would be sent are your custom attributes).
```
rchinn-segment, line 121:
```suggestion
4. *Event Properties* is set from the path `properties` by default, which means that all the properties that you include in your Segment `analytics.track` call are passed to Friendbuy. If you only want to pass custom attributes to Friendbuy, you need to segregate the fields to pass to Friendbuy into an object and then specify the path to that object, similar to what is described above in [Custom Attributes](#custom-attributes).
```

rchinn-segment (changes requested):
I've added a few edits and left some suggested changes for review.

PR #2286, opened 2021-12-15
Add note to hubspot docs about event tester

kdaswani:
### Proposed changes
We've been getting a lot of tickets about the Hubspot event tester not working which is expected due to how identify calls and group calls need to be sent together and in order, otherwise group calls fail. A docs update will help clarify this to customers.

### Merge timing
- ASAP once approved

### Related issues (optional)
https://segment.atlassian.net/browse/DEST-3601

### DIFF
--- a/src/connections/destinations/catalog/hubspot/index.md
+++ b/src/connections/destinations/catalog/hubspot/index.md
@@ -8,7 +8,7 @@ cmode-override: true
 [HubSpot](https://www.hubspot.com/) is an inbound marketing and sales platform that helps companies attract visitors, convert leads, and close customers. The `analytics.js` HubSpot Destination is open-source. You can browse the code [on GitHub](https://github.com/segmentio/analytics.js-integrations/tree/master/integrations/hubspot).
 
 > warning ""
-> The HubSpot destination does not work with Segment's Event Tester due to how identify and group calls must be sent in order. The Event Tester should not be used for troubleshooting the HubSpot destination.
+> The HubSpot destination is not compatible with the Segment Event Tester. As result, Segment recommends using other tools to troubleshoot the HubSpot destination.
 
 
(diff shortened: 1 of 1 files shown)


src/connections/destinations/catalog/hubspot/index.md
```diff
@@ -7,6 +7,9 @@ cmode-override: true
 
 [HubSpot](https://www.hubspot.com/) is an inbound marketing and sales platform that helps companies attract visitors, convert leads, and close customers. The `analytics.js` HubSpot Destination is open-source. You can browse the code [on GitHub](https://github.com/segmentio/analytics.js-integrations/tree/master/integrations/hubspot).
 
+> warning ""
+> The HubSpot destination does not work with Segment's Event Tester due to how identify and group calls must be sent in order. The Event Tester should not be used for troubleshooting the HubSpot destination.
```
markzegarelli, line 11:
```suggestion
> The HubSpot destination is not compatible with the Segment Event Tester. The HubSpot destination does not send Identify and Group calls in order, as required by the Event tester. As result, Segment recommends using other tools to troubleshoot the HubSpot destination.
```

markzegarelli (changes requested):
Just some rewording of the note.

src/connections/destinations/catalog/hubspot/index.md
kdaswani: Thanks! I think this line is incorrect `The HubSpot destination does not send Identify and Group calls in order, as required by the Event tester.` It's the opposite. Our HubSpot destination DOES send those in order, but the Event Tester cannot chain two events (you just send one at a time, so if you try to send a Group call it fails). Maybe we remove this detail.

markzegarelli (approved):

PR #2457, opened 2022-02-03
remove salesforce DMP from salesforce list

kdaswani:
### Proposed changes
Salesforce DMP is in private beta, and was built 4+ years ago for Fox specifically. No customers should be using this destination, and StratConn is investigating deprecating it. In the meantime, I would like to remove it from the Salesforce nav in our docs, so customers cannot find the doc. It is already hidden but is still appear in the left-hand column:

Screen Shot 2022-02-03 at 1 17 50 PM

I think this is because the strat = salesforce. If there's another reason it is still appearing, please update this PR to remove that. I don't see it when I'm searching so I think it is unindexed already.

### Merge timing
- ASAP once approved

### DIFF
--- a/src/_data/sidenav/strat.yml
+++ b/src/_data/sidenav/strat.yml
@@ -80,8 +80,8 @@ sections:
   #   title: Salesforce Live Agent destination
   - path: /connections/destinations/catalog/salesforce-marketing-cloud
     title: Salesforce Marketing Cloud destination
-  - path: /connections/destinations/catalog/salesforce-dmp
-    title: Salesforce DMP destination (beta)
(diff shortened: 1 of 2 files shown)


markzegarelli:
@kdaswani Should we just unpublish this completely?

kdaswani:
Good question! Let's keep it unindexed for now just in case Fox still needs to reference it. In my JIRA to eventually deprecate, I've added a note to make sure we unpublish upon deprecation: https://segment.atlassian.net/browse/STRATCONN-1151. Thank you!

markzegarelli (approved):

PR #2520, opened 2022-02-18
OneSignal update DOC-464

pwseg:
### Proposed changes

Added note about OneSignal Data Lag limits.

### Merge timing

ASAP once approved.

### DIFF
--- a/src/connections/destinations/catalog/onesignal-new/index.md
+++ b/src/connections/destinations/catalog/onesignal-new/index.md
@@ -48,6 +48,9 @@ Segment sends Identify traits as [Player Data Tags](https://documentation.onesig
 
 ![](images/demo.png)
 
+> info "Data Tag Limits"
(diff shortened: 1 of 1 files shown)


src/connections/destinations/catalog/onesignal-new/index.md
```diff
@@ -48,6 +48,9 @@ Segment sends Identify traits as [Player Data Tags](https://documentation.onesig
 
 ![](images/demo.png)
 
+> info "Data Tag Limits"
+> Your OneSignal plan may cap incoming Data Tags. If you notice successful Segment events that don't update in OneSignal, you've likely reached your OneSignal plan Data Tag limit.
```
stayseesong, line 52: @pwseg I took a look at the Jira ticket, and I think it should say that if the user traits aren't updating, not the events. Maybe something like this?

Your OneSignal plan may cap incoming Data Tags. Once you've reached your Data Tag limit, your user traits won't update in your OneSignal destination.

pwseg: Great catch.  Done!

stayseesong (approved):

PR #2531, opened 2022-02-23
Update Add Destinations doc

tawe-segment:
### Proposed changes
Update docs to state that we do not support connecting multiple sources to the same instance of a destination.

### Merge timing
- ASAP once approved

### DIFF
--- a/src/connections/destinations/add-destination.md
+++ b/src/connections/destinations/add-destination.md
@@ -114,6 +114,12 @@ Some destinations do not support having multiple instances connected to the same
 
 You can create unique destination filters for each destination instance connected to the same source.
 
+
+### Connect multiple sources to one instance of a destination
+
+It is not possible to connect more than one source to the same instance of a destination. Instead, you can create another instance of the same destination for the other sources, and click **Copy Settings From Other Destination** to save yourself time entering the settings values again.
+
+
 ### Connect to more than one instance of a destination using the Config API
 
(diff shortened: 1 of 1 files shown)


src/connections/destinations/add-destination.md
```diff
@@ -114,6 +114,12 @@ Some destinations do not support having multiple instances connected to the same
 
 You can create unique destination filters for each destination instance connected to the same source.
 
+
+### Connect multiple sources to one instance of a destination
+
+It is not possible to connect more than one source to the same instance of a destination. However, you can create another instance of the same destination for the other sources, and click **Copy Settings From Other Destination** to save yourself time entering the settings values again.
```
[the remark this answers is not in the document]
tawe-segment: @forstisabella -  correct, if you create two separate website sources, each with different writekeys, you cannot point them both to the same instance of a destination without having to create that destination twice for each individually.

So for your example with two sources:
Site A - writekey: 123
Site B - writekey 789

You will have to create 2 HubSpot instances, one connected to the Site A source and another to the Site B source = 2 instances of HubSpot.

And that will look something like this on the Destinations UI even though they are all sending data to the same HubSpot ID:
image

forstisabella, line 120: Okay that makes sense, thank you for the clarification! Please let me know if the following sentence sounds right:

```suggestion
It is not possible to connect multiple instances of one source (for example, two website sources) to the same destination. However, you can create another instance of the destination for the other sources, and click **Copy Settings From Other Destination**  to save yourself time entering the settings values again.
```

tawe-segment: lgtm! thanks!

forstisabella (approved):
Looks great!

PR #2534, opened 2022-02-23
Batch-audience-compute-clarification

rchinn-segment:
### Proposed changes
Added a compute time clarification for account-level audiences. Based on DOC-465. 

### Merge timing

ASAP once approved. 
### Related issues (optional)
DOC-465

### DIFF
--- a/src/personas/audiences/account-audiences.md
+++ b/src/personas/audiences/account-audiences.md
@@ -100,7 +100,7 @@ For example, you may wish to create an audience which selects all admin-level us
 
 ## Known limitations of account-level audiences
 
-- Unlike user-level audiences, which are [computed in real time](/docs/personas/audiences#realtime-compute-vs-batch), account-level audiences are computed on a batched basis. Segment computes account-level audiences roughly every hour, but it's important to note that compute times might fluctuate.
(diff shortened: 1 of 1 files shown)


src/personas/audiences/account-audiences.md
```diff
@@ -100,7 +100,7 @@ For example, you may wish to create an audience which selects all admin-level us
 
 ## Known limitations of account-level audiences
 
-- Unlike user-level audiences which are [computed in real time](/docs/personas/audiences#realtime-compute-vs-batch), account-level audiences are computed on an hourly, batched basis.
+- Unlike user-level audiences, which are [computed in real time](/docs/personas/audiences#realtime-compute-vs-batch), account-level audiences are computed on a batched basis. Segment computes account-level audiences roughly every hour, but it's important to note that compute times might fluctuate.
```
stayseesong, line 103:
```suggestion
- Unlike user-level audiences, which are [computed in real time](/docs/personas/audiences#realtime-compute-vs-batch), account-level audiences are computed on a batched basis. Segment computes account-level audiences roughly every hour, but it's important to note that compute times can fluctuate based on the load on the system.
```

stayseesong: @rchinn-segment I would also share the change with the reporter of the Jira ticket to make sure this captures their request and see if they have other suggestions. Usually I provide a screenshot of the change and ask the reporter if it's ok.

rchinn-segment: @stayseesong Great suggestions, thanks!

stayseesong (approved):

PR #2548, opened 2022-03-01
Describe the required Friendbuy web destination fields.

dhull:
This is a draft request and I will be soliciting Friendbuy feedback before I make it official.

### Proposed changes

Added section describing the required fields for the Friendbuy web destination and the consequence if the required fields are not populated in the analytics calls. We have merchants who are attempting to enable the Friendbuy web destination whose data is not being sent to Friendbuy because they are not populating the required fields.

Currently the list of fields generated by `{% include components/actions-fields.html %}` does not call out the required fields. It would be really cool if it did.

### Merge timing

ASAP once approved

### DIFF
--- a/src/connections/destinations/catalog/actions-friendbuy/index.md
+++ b/src/connections/destinations/catalog/actions-friendbuy/index.md
@@ -51,7 +51,21 @@ Before you start, you must have Segment's Analytics.js 2.0 installed on your sit
 
 {% include components/actions-fields.html %}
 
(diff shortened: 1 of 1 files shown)


markzegarelli:
Hi @dhull, thanks for calling this out. I do see that we receive a `required` flag in our destination metadata, so I can add a required indicator to the individual tables directly.

markzegarelli:
Closing this PR since I can't update the external branch. 

I've added an indicator of required status directly in the definitions table in PR #2559 . 

Screen Shot 2022-03-03 at 10 44 53 AM

PR #2615, opened 2022-03-17
[index.md] update Bucket destination copy

tcgilbert:
### Proposed changes

Updated bucket copy per partner request. 

There is one part of the docs that they wanted to update, that is not included in the index.md file. 

The description in the settings display:
Screen Shot 2022-03-17 at 4 59 08 PM

They want it updated from:

- string. The API key is your App ID. Find it under Settings (cog wheel) when logged in.

To 

+ string. The API key is your app’s Tracking Key. Locate it under Settings when signed in.

This is now [mapped correctly in the partner portal](https://app.segment.com/partner-portal/integration/bucket/settings), so I am guessing this will be included in the next build?

### Merge timing
asap

### DIFF
--- a/src/connections/destinations/catalog/bucket/index.md
+++ b/src/connections/destinations/catalog/bucket/index.md
@@ -3,9 +3,10 @@ title: Bucket Destination
 rewrite: true
 id: 5fabc0b00f88248bbce4db48
 ---
-[Bucket](https://bucket.so/?utm_source=segmentio&utm_medium=docs&utm_campaign=partners) features analytics for SaaS companies.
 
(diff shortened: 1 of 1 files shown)


rchinn-segment (commented):
Thanks for adding this @tcgilbert! The Partner Portal content is mapped correctly, but I noticed “Settings” is missing the “s”. This needs to be corrected in the portal, and then we can include this update in our next docs release.

tcgilbert:
Nice catch @rchinn-segment! Just updated partner portal

rchinn-segment (approved):

PR #2704, opened 2022-04-05
Add UA to GA4 section

kdaswani:
### Proposed changes
We've been getting a lot of questions and appetite for general GA4 best practices, especially in light of Universal Analytics being deprecated next summer. This section attempts to answer some of those questions. There may be opportunity to add to this section, but this is a good starting place.

There's a lot of content and links added so would love a thorough review to make sure I didn't mess anything up 😄 

### Merge timing
- ASAP once approved

### Related issues (optional)
JIRA: https://segment.atlassian.net/browse/STRATCONN-1279


kdaswani:
Out of curiosity, is there a way to hide the fields within each Action by default? When they are fully expanded, the page becomes very long because there are tons of actions. Would love if we could keep the fields collapsed like this:
Screen Shot 2022-04-04 at 5 05 40 PM

markzegarelli:
> Out of curiosity, is there a way to hide the fields within each Action by default? When they are fully expanded, the page becomes very long because there are tons of actions. Would love if we could keep the fields collapsed like this: Screen Shot 2022-04-04 at 5 05 40 PM


We used to have them collapsed by default, but we got a request to have them open because people were missing the button. Let me have a think about this to figure out a better solution.

rchinn-segment (approved):

PR #2805, opened 2022-04-27
Bundle Obfuscation [DOC-508]

arielsilvestri:
### Proposed changes

Announcing our obfuscation feature to customers.

### Merge timing

Need to clarify if we have any other comms to make before merging this, but I will ping here once I have that confirmed.

### DIFF
--- a/src/connections/sources/catalog/libraries/website/javascript/index.md
+++ b/src/connections/sources/catalog/libraries/website/javascript/index.md
@@ -493,8 +493,6 @@ An example:
 analytics.load('writekey', { integrations: { All: false, 'Google Analytics': true, 'Segment.io': true } })
 ```
 
-<<<<<<< Updated upstream
-=======
 You can also add an `obfuscate` property to the object taken as a second argument, which will obscure the url from which your integrations and destination actions are loaded. This is helpful for names that are commonly flagged by ad blockers.
 
 An example:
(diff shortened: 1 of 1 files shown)


src/connections/sources/catalog/libraries/website/javascript/index.md
```diff
@@ -493,6 +493,16 @@ An example:
 analytics.load('writekey', { integrations: { All: false, 'Google Analytics': true, 'Segment.io': true } })
 ```
 
+You can also add an `obfuscate` property to the object taken as a second argument, which will obscure the url from which your integrations and actions are loaded. This is helpful for names that are commonly flagged by ad blockers.
```
stayseesong, line 496: @arielsilvestri by actions, do you mean destination actions?

arielsilvestri: 1. Correct, I will update that to say "integrations and destination actions".
2. Sure, some clarification there - the goal of [this project was to provide ways for our customers to help mitigate how often commonly blocked words in urls would prevent certain integrations from properly loading. With this change, something like `adobe-analytics` would no longer appear in any part of the URL, as a small example.

stayseesong:
hi @arielsilvestri was your most recent commit intentional? Looks like the rebase you committed didn't resolve conflicts properly

arielsilvestri:
ack, weird. Let me see if I can fix that.

stayseesong (approved):

PR #2815, opened 2022-04-28
Flag that cloud mode destinations are still supported

niallzato:
### Proposed changes
Hey @markzegarelli  get your thoughts on this? I think there's bit of a disconnect with customers on what is supported for our new mobile SDKs. I want to flag that the integrations we list are device mode, but all historical cloud destinations are still supported.


markzegarelli (approved):

markzegarelli:
LGTM, @niallzato. Thanks!

PR #2849, opened 2022-05-06
Add 'in' operator to FQL documentation

tysonmote:
### Proposed changes

This PR adds the new 'in' operator to the FQL documentation.

### Merge timing

As soon as it is approved.


src/config-api/fql.md
```diff
@@ -100,6 +105,7 @@ If your field name has a character not in the set of `{a-z A-Z 0-9 _ -}`, you mu
 | `>=`     | `number`                                      | `number`                                      | `true` if the left side is greater than or equal to the right side.                                         |
 | `<`      | `number`                                      | `number`                                      | `true` if the left side is less than the right side.                                                        |
 | `<=`     | `number`                                      | `number`                                      | `true` if the left side is less than or equal to the right side.                                            |
+| `in`     | `string`, `number`, `bool`, or `null`         | `list`                                        | `true` if the left side is contained in the list of values.                                                 |
```
jayakrishnannair, line 108: Looks good. Wondering if we should mention that using `in` is much more efficient than chaining `or` conditions.

jayakrishnannair (approved):

src/config-api/fql.md
tysonmote: I don't think so, personally. Customers shouldn't have to worry about our internal implementation's efficiency, and it wouldn't make a difference to them anyways.

rchinn-segment (approved):

PR #2917, opened 2022-05-18
Update documentation about testing

g-normand:
### Proposed changes

I add some informations about not sending to Segment while testing

### DIFF
--- a/src/connections/sources/catalog/libraries/server/python/index.md
+++ b/src/connections/sources/catalog/libraries/server/python/index.md
@@ -45,6 +45,12 @@ analytics.debug = True
 analytics.on_error = on_error
 ```
 
+If you don't want to send Segments informations while testing, you need to add in your test:
+
+```python
+analytics.send = False
+```
+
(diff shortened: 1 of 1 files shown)


src/connections/sources/catalog/libraries/server/python/index.md
```diff
@@ -45,6 +45,12 @@ analytics.debug = True
 analytics.on_error = on_error
 ```
 
+If you don't want to send Segments informations while testing, you need to add in your test:
```
markzegarelli, line 48:
```suggestion
If you don't want to send information to Segment during testing, add the following code to your test:
```

markzegarelli (changes requested):

src/connections/sources/catalog/libraries/server/python/index.md
g-normand: Sorry, i'm not used to  github PR interface. Thanks for the review :)

markzegarelli (approved):

PR #2942, opened 2022-05-24
Variance alias support

gmathieu:
### Proposed changes

Add Alias method docs now that Variance supports it.

### Merge timing

ASAP 🙏

### DIFF
--- a/src/connections/destinations/catalog/variance/index.md
+++ b/src/connections/destinations/catalog/variance/index.md
@@ -85,6 +85,16 @@ analytics.group("groupId123", {
 
 Segment sends Group calls to Variance as an `Account` if you've chosen the "Group" Account Mapping method during setup.
 
+### Alias
+
+If you aren’t familiar with the Segment Spec, take a look at the [Alias method documentation](/docs/connections/spec/alias/) to learn about what it does. An example call would look like this:
+
+```js
+analytics.alias('123456')
+```
+
+If there is an existing user with that `userId` in Variance that matches the `previousId` passed with the alias that user will be merged into the "123456" user identified in the alias. If there is only one user with the `previousId` and no user with the `userId` that user will simply have their Variance `externalId` updated to match the current `userId` passed in the alias.
+
 ## Account Mapping
 
 Variance offers several ways to map your users to accounts or companies, including the following.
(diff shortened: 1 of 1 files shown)


gmathieu:
Hi @stayseesong,

I'm waiting on Segment's engineering team to update our integration. Once updated, I'll submit this PR for review. Here's the [ticket](https://segment.zendesk.com/hc/requests/473200) if you have access.

Thanks for following up

src/connections/destinations/catalog/variance/index.md
```diff
@@ -85,6 +85,16 @@ analytics.group("groupId123", {
 
 Segment sends Group calls to Variance as an `Account` if you've chosen the "Group" Account Mapping method during setup.
 
+### Alias
+
+If you aren’t familiar with the Segment Spec, take a look at the [Alias method documentation](/docs/connections/spec/alias/) to learn about what it does. An example call would look like this:
```
stayseesong, line 90:
```suggestion
If you're not familiar with the Segment Spec, take a look at the [Alias method documentation](/docs/connections/spec/alias/) to learn about what it does. An example call would look like this:
```

src/connections/destinations/catalog/variance/index.md
```diff
@@ -85,6 +85,16 @@ analytics.group("groupId123", {
...
+
+```js
+analytics.alias('123456')
+```
+
+If there is an existing user with that `userId` in Variance that matches the `previousId` passed with the alias that user will be merged into the "123456" user identified in the alias. If there is only one user with the `previousId` and no user with the `userId` that user will simply have their Variance `externalId` updated to match the current `userId` passed in the alias.
```
stayseesong, line 96:
```suggestion
If there is an existing user with that `userId` in Variance that matches the `previousId` passed with the alias, that user will be merged into the "123456" user identified in the alias. If there is only one user with the `previousId` and no user with the `userId`, that user will have their Variance `externalId` updated to match the current `userId` passed in the alias.
```

stayseesong (dismissed):

gmathieu:
@stayseesong I addressed your feedback. Thanks!

markzegarelli (approved):

PR #3009, opened 2022-06-09
Deprecate Dev Center 1.0

niallzato:
### Proposed changes
Flag that Dev Center 1.0 is no longer accepting new submissions.


forstisabella (approved):
I added an attribute to the Airtable links so they'll open in a new tab, but the content looks good to me!

PR #3044, opened 2022-06-14
Remove Databrains destination

charholt:
### Proposed changes
Customer is no longer in business and integration is now unavailable to customers

### Merge timing
- ASAP once approved?


markzegarelli (approved):

markzegarelli:
Thanks, @charholt! I noticed that this is also listed as `unavailable` in the API, so we're good to go here.

PR #3083, opened 2022-06-21
Remove user agent for ios 

niallzato:
### Proposed changes
User agent is not collected in analytics-ios. Updating docs to reflect this.


src/connections/spec/common.md
```diff
@@ -276,7 +276,7 @@ Other libraries only collect `context.library`, any other context variables must
 | screen.height            |              | ✅             | ✅                 |
 | screen.width             |              | ✅             | ✅                 |
 | traits                   |              | ✅             | ✅                 |
-| userAgent                | ✅            | ✅             | ✅                 |
+| userAgent                | ✅            |              | ✅                 |
 | timezone                 |              | ✅             | ✅                 |
 
 - IP Address is not collected by Segment's libraries, but instead filled in by Segmen'ts servers when it receives a message for **client side events only**.
```
stayseesong, line 282:
```suggestion
- IP Address isn't collected by Segment's libraries, but instead filled in by Segment's servers when it receives a message for **client side events only**.
```

stayseesong (approved):

PR #3095, opened 2022-06-22
Add Klaviyo Personas Guidance #2755

### LINKED ISSUE #2755: [Content]: Klaviyo Personas Section

https://segment.com/docs/connections/destinations/catalog/klaviyo/#personas

The existing Personas content for the Klaviyo destination doc does a nice job explaining our standard-ish practice of sending identify() calls downstream, which update user profiles in Klaviyo. But it's missing two things that I think make things misleading for users:

1. The following is _implied_ but I think should be explicitly stated at the end of the Personas section: "Once Segment's identify() call sets a custom property on the Klaviyo user profile for a Personas audience, Klaviyo segments can be created off that property when its value is "true". This will ensure membership is updated automatically as users fall in and out of eligibility criteria for the Personas audience". 
2. In the Segment UI, a user is prompted to enter a list_id as an optional field. If that's filled in, our [Klaviyo destination code](https://github.com/segmentio/integrations/blob/24fd18d9fe74a47512558172614be71be8b29fd7/integrations/klaviyo/lib/index.js#L63) will fire its _addToList method instead of the normal identify() approach. _addToList, from what I can tell, only ADDS users to the provided list_id. It does not provide a way to REMOVE users. So by including a list_id in the destination settings, it breaks the expectation set by the existing Personas language (which implies Personas will automatically keep the membership up to date). So in tandem with (1) above, I think the writeup should include a warning box for Personas along the lines of "inputting a list id in the destination settings is not recommended for Personas destinations. Doing so will instruct Segment to add users to the provided list_id, but users will not be removed. Instead leave list id empty and see guidance above for creating Klaviyo segments" (and reference (1)). 

### Additional information

_No response_

pwseg:
### Proposed changes

- Added guidance on leaving out `list_id` when setting up the Klaviyo Destination.
- Specified JavaScript next to Markdown code blocks to enable syntax highlighting.
- Added `{target="_blank"}` to external links.
- Changed heading case to sentence case.

### Merge timing

- ASAP once approved.

### Related issues (optional)

- Closes #2755.

### DIFF
--- a/src/connections/destinations/catalog/klaviyo/index.md
+++ b/src/connections/destinations/catalog/klaviyo/index.md
@@ -4,21 +4,21 @@ rewrite: true
 cmode-override: true
 id: 54521fd825e721e32a72eec8
 ---
-[Klaviyo](https://www.klaviyo.com/features/overview) is a powerful email platform focused on ecommerce that helps companies make more money. It supports segmentation based on category and event triggers like product bought, page viewed, email engagement, or amount spent.
+[Klaviyo](https://www.klaviyo.com/features/overview){:target="_blank"} is a powerful email platform focused on ecommerce that helps companies make more money. It supports segmentation based on category and event triggers like product bought, page viewed, email engagement, or amount spent.
 
 It measures opens, clicks, revenue generated, breakdown of generated revenue based on custom attributes (like campaign type or amount gained per recipient), and provides trend reports, cohort analysis, and subscriber growth
 
-Ultimately, Klaviyo lets you send personalized newsletters, automates triggered emails, product recommendations, welcome campaigns, order announcements, push notifications and sync your data to Facebook custom audiences.
+Klaviyo lets you send personalized newsletters, automates triggered emails, product recommendations, welcome campaigns, order announcements, push notifications and sync your data to Facebook custom audiences.
 
 To configure Klaviyo as an Event Source to get data into your warehouse or other downstream tools, see the [Klaviyo Source](/docs/connections/sources/catalog/cloud-apps/klaviyo/) documentation.
 
-## Getting Started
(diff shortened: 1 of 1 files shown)


src/connections/destinations/catalog/klaviyo/index.md
```diff
@@ -2,23 +2,24 @@
...
-## Getting Started
+## Getting started
 
 {% include content/connection-modes.md %}
 
 1. From the Segment web app, click **Catalog**.
 2. Search for "Klaviyo" in the Catalog, select it, and choose which of your sources to connect the destination to.
-3. Navigate to your [Account > Settings > API Keys](https://www.klaviyo.com/account#api-keys-tab){:target="_blank"} in the Klaviyo's UI and copy your "API Key" into the Segment Settings UI.
+3. Navigate to [Account > Settings > API Keys](https://www.klaviyo.com/account#api-keys-tab){:target="_blank"} in Klaviyo's UI and copy your API Key into the Segment Settings UI.
 4. **Note:** Klaviyo requires the Private API Key to use the List API. You can find this by going to Klaviyo's UI and clicking [Account > Settings > API Keys > Create API Key](https://www.klaviyo.com/account#api-keys-tab){:target="_blank"}  to generate a Private API Key and copy it into the Segment Settings UI.
```
stayseesong, line 23:
```suggestion
> info ""
> Klaviyo requires the Private API Key to use the List API. You can find this by going to Klaviyo's UI and clicking [Account > Settings > API Keys > Create API Key](https://www.klaviyo.com/account#api-keys-tab){:target="_blank"}  to generate a Private API Key and copy it into the Segment Settings UI.
```

stayseesong (approved):

PR #3324, opened 2022-08-03
Add alias for Swift docs

niallzato:
### Proposed changes
A customer wrote in that we do not list the Alias method in our docs. We do mention it in the migration guide, but we are lacking a definition in the main docs: https://segment.com/docs/connections/sources/catalog/libraries/mobile/swift-ios/migration/

### DIFF
--- a/src/connections/sources/catalog/libraries/mobile/swift-ios/index.md
+++ b/src/connections/sources/catalog/libraries/mobile/swift-ios/index.md
@@ -34,7 +34,7 @@ To get started with the Analytics-Swift mobile library:
 
     ```swift
     var analytics: Analytics? = nil
-    
+
     func application(_ application: UIApplication, didFinishLaunchingWithOptions launchOptions: [UIApplication.LaunchOptionsKey: Any]?) -> Bool {
             // Override point for customization after application launch.
             let configuration = Configuration(writeKey: "WRITE_KEY")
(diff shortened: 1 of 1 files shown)


src/connections/sources/catalog/libraries/mobile/swift-ios/index.md
```diff
@@ -186,6 +186,22 @@ analytics.group(groupId: "user-123", traits: MyTraits(
 {% endcodeexampletab %}
 {% endcodeexample %}
 
+### Alias
+The [Alias](/docs/connections/spec/alias/) method is used to merge two user identities, effectively connecting two sets of user data as one. When this method is called the `newId` value will overwrite the old userId. If no userId is currently set, the `newId` will be associated with future events as the userId. This is an advanced method and may not be supported across the entire destination catalog, please refer to the destination docs to understand if this method is supported or not.
```
stayseesong, line 190:
```suggestion
The [Alias](/docs/connections/spec/alias/) method is used to merge two user identities, effectively connecting two sets of user data as one. When this method is called, the `newId` value overwrites the old `userId`. If no `userId` is currently set, the `newId` associates with future events as the `userId`. This is an advanced method and may not be supported across the entire destination catalog. 
```

stayseesong (changes requested):

stayseesong (approved):

stayseesong:
@niallzato I edited the description of alias a bit, lmk what you think

niallzato:
> 

LGTM ! @stayseesong  you good to merge this?

PR #3331, opened 2022-08-04
Update copy in Google Tag Manager article

vanand17:
### Proposed changes

Changing a letter from upper case to lower case based on how the event shows up in the destination.

### Merge timing

When approved.

### DIFF
--- a/src/connections/destinations/catalog/google-tag-manager/index.md
+++ b/src/connections/destinations/catalog/google-tag-manager/index.md
@@ -33,7 +33,7 @@ analytics.page('Home', {
 You must call the Page method for Google Tag Manager to load. Segment includes a call to `analytics.page` in your default Segment snippet, so if you haven't removed that, GTM will work the same as if you installed the GTM snippet directly.
 
 ### Tracking All Pages
-When you turn on the setting to **Track All Pages** in your Optional Settings, Segment tracks events whenever you call the `page` method and send a "Loaded A Page" event to Google Tag Manager. See the `track` section below for more info on how Segment sends events to GTM.
+When you turn on the setting to **Track All Pages** in your Optional Settings, Segment tracks events whenever you call the `page` method and send a "Loaded a Page" event to Google Tag Manager. See the `track` section below for more info on how Segment sends events to GTM.
 
 ### Named Page Events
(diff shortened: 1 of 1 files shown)


kdaswani:
@vanand17 Have we confirmed this with a live test? I actually don't see anywhere in the code where we send "Loaded a Page" so want to make sure this is actually all true. https://github.com/segmentio/analytics.js-integrations/blob/master/integrations/google-tag-manager/lib/index.js

Probably happens somewhere else but would love 100% confirmation from a true test if possible. We should also probably update the setting description to lowercase "a" in partner portal as well if so: https://app.segment.com/partner-portal/integration/google-tag-manager/settings

pwseg (approved):

kdaswani:
@vanand17 following up on my last comment before we merge this in.

> Have we confirmed this with a live test? I actually don't see anywhere in the code where we send "Loaded a Page" so want to make sure this is actually all true. https://github.com/segmentio/analytics.js-integrations/blob/master/integrations/google-tag-manager/lib/index.js
> 
> Probably happens somewhere else but would love 100% confirmation from a true test if possible. We should also probably update the setting description to lowercase "a" in partner portal as well if so: https://app.segment.com/partner-portal/integration/google-tag-manager/settings

vanand17:
Hi Kiara,

I saw this request being approved hence I did not go ahead and run a test
myself. Just to probably clarify, the change that was requested by a
customer was around a spelling error which is currently reading 'Loaded A
Page' and this could be changed to 'Loaded a Page' which a lower case 'a'.
Attaching a screenshot below for your reference.

[image: image.png]
[image: image.png]

Link to the sections of our docs -
https://segment.com/docs/connections/destinations/catalog/google-tag-manager/#tracking-all-pages

https://segment.com/docs/connections/destinations/catalog/google-tag-manager/#settings


Apologies if this was not clear, please let me know if you need any further
clarification :)

kdaswani:
Hi @vanand17 - thanks! I ran a test myself to be 100% sure that is what we send. I see that upon a page call, if tracking all pages, we see "Loaded a Page" in the GTM data layer. Approving this PR given my test. I will update the setting description in partner portal to also lowercase the "a" to match
Screen Shot 2022-08-24 at 10 24 55 AM
.

kdaswani (approved):

vanand17:
Thanks Kiara.

PR #3361, opened 2022-08-11
Add and hide new Pardot (Actions) Docs

kdaswani:
### Proposed changes
StratConn just rebuilt Pardot on Actions! This is our first draft of the documentation for it. We plan to go to Private Beta on August 23 or 24 with a couple of customers. Would love to have these docs on segment.com/docs but hidden and unindexed so we can share the URL with beta customers.

The destination is currently in Private Building but we will move it to Private Beta ahead of August 23. @markzegarelli and team - is it possible to get the autogenerated tables (i.e. settings, available actions, dossier) included once in Private Beta, or does it need to be Public Beta for that to work? I've included the `{% include components/actions-fields.html %}` placeholder for now.

### Merge timing
- Happy to publish ASAP as hidden so we can confirm everything looks good. I can file another PR ahead of Private Beta if we need any changes or if it's required to help autogenerate the tables once the status in partner portal has been flipped.

Thank you!
cc @kwang122 @marinhero


stayseesong:
@kdaswani the autogenerated tables currently don't generate right now because the destination is in private building and the status hasn't been flipped in partner portal. With that said, I've hidden the dossier and also commented out the `{% include components/actions-fields.html %}` because nothing gets generated right now and it just shows up as empty. We'll definitely need to add those back in once the status changes in partner portal

stayseesong (approved):

kdaswani:
Sounds good thank you @stayseesong! Can you confirm this - will the tables be autogenerated if the status in partner portal is "Private Beta" or does it need to be "Public Beta"? Ideally, we could get the tables generated upon Private Beta if possible. I can file a separate PR to remove the commented out piece and add the dossier once we flip to "Private Beta" if so. Thanks again!

PR #3395, opened 2022-08-23
Actions moengage

aekanshd:
### Proposed changes

Added update docs for MoEngage (Actions) Destination Integration.

### Merge timing
ASAP once approved.


aekanshd:
Can you also help us understand how to populate the “old vs new” mapping table? It is given by this command in the document: `{% include components/actions-map-table.html name="moengage" %}`. When I run the dev server, this table is not populated. We are already in Private Beta phase. I only see this in `localhost` dev server. I have added the config API tokens also. This is what I see:

image

markzegarelli:
@aekanshd That table requires manual input. If you have a mapping of MoEngage Classic to MoEngage (Actions) settings, I can assist with setting that up. Unfortunately, there is no way for the app to understand how to mirror functionality between the two versions of the destination.

If you don't have this yet, we can hide this section for now.

markzegarelli:
Closing. This change was deployed in #3397

aekanshd:
Was there a reason the table and the image were omitted from the final doc? Refer to this commit: https://github.com/segmentio/segment-docs/pull/3397/commits/23da57fe3358c6db701b697913550cf6a1840361

PR #3409, opened 2022-08-25
Add autogenerated tables and dossier back to Pardot (Actions)

kdaswani:
### Proposed changes
Hi team! We just moved `actions-pardot` to Private Beta. My understanding from @stayseesong is that this will now allow the autogenerated components and dossier to populate automatically. I have uncommented those tables out. Please let me know if this is incorrect, but we are hoping to get those added in so we can start sharing the docs with our private beta customers later this week/early next.

### Merge timing
- ASAP once approved (Thurs Aug 25 deploy if possible!)


stayseesong:
@kdaswani the autogenerated tables will show up after merging and deploying this PR, but to get the dossier to show up, we'll need to update that for the next deploy on Tuesday. Hope that's ok!

stayseesong (approved):

PR #3416, opened 2022-08-25
Update Snapchat Conversions API docs prior to GA

kdaswani:
### Proposed changes
We are bringing Snapchat Conversions API to GA next Tues, 8/30! These updates:

- Remove the beta notice
- Add a note to the FAQs about double hashing, as surfaced in our Launch Council

We will move the partner portal status to Public EOD Monday or early Tues AM before the docs deploy so that the dossier is also updated. Thanks!

### Merge timing
- On a specific date: Tues 8/30 docs deploy

### Related issues (optional)
https://segment.atlassian.net/browse/STRATCONN-1476

### DIFF
--- a/src/connections/destinations/catalog/actions-snap-conversions/index.md
+++ b/src/connections/destinations/catalog/actions-snap-conversions/index.md
@@ -81,4 +81,4 @@ In addition, Segment creates a SHA-256 hash of the following fields before sendi
 - IP Address
 
 > warning ""
-> If you hash identifiers upstream before sending to Segment, Segment will still hash that data before sending to Snap. This results in a double hash that will not be able to be matched on. Please ensure your fields are not hashed prior to sending through the Snapchat Conversions API destination.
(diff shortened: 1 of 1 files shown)


src/connections/destinations/catalog/actions-snap-conversions/index.md
```diff
@@ -83,3 +80,5 @@ In addition, Segment creates a SHA-256 hash of the following fields before sendi
 - Phone Number
 - IP Address
 
+> warning ""
+> If you hash identifiers upstream before sending to Segment, Segment will still hash that data before sending to Snap. This results in a double hash that will not be able to be matched on. Please ensure your fields are not hashed prior to sending through the Snapchat Conversions API destination.
```
stayseesong, line 84:
```suggestion
> If you hash identifiers upstream before sending to Segment, Segment still hashes that data before sending to Snap. This results in a double hash that won't be able to be matched on. Please ensure your fields are not hashed prior to sending through the Snapchat Conversions API destination.
```

stayseesong (approved):

kdaswani:
Can we merge this one in for tomorrow's deploy @stayseesong? I'll move the partner portal status to `Public` tonight so the status in the dossier is updated upon deploy!

PR #3444, opened 2022-09-02
Customer.io destination actions don't support alias or group calls

glosier:
### Proposed changes

Remove `alias` and `group` methods from Customer.io Destination Actions page. 

I had some local environment issues and struggled to build locally, but I'm pretty sure I made this change in the right place.

### Merge timing
- ASAP once approved

### DIFF
--- a/src/_data/catalog/destinations.yml
+++ b/src/_data/catalog/destinations.yml
@@ -12673,8 +12673,8 @@ items:
   methods:
     track: true
     identify: true
-    group: true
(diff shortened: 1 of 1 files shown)


markzegarelli:
Hi @glosier. These updates need to be made in the back end by our Support team. Can you open a ticket with them?

It's a pretty easy change and should not take much time at all.

Closing this PR for now, as this change will be overwritten with our next catalog update.

PR #3484, opened 2022-09-13
Add info for uuid during local development

niallzato:
### Proposed changes
Due to limitations with React Native, we do not set UUID during local development. UUID will be set on a live app for anonymousId and messageId, but not locally.

### DIFF
--- a/src/connections/sources/catalog/libraries/mobile/react-native/index.md
+++ b/src/connections/sources/catalog/libraries/mobile/react-native/index.md
@@ -511,7 +511,7 @@ No, only the plugins listed above are supported in device-mode for Analytics Rea
 ### Will I still see device-mode integrations listed as `false` in the integrations object?
 When you successfully package a plugin in device-mode, you won't see the integration listed as `false` in the integrations object for a Segment event. This logic is packaged in the event metadata, and isn't surfaced in the Segment debugger.
 ### Why are my IDs not set in UUID format?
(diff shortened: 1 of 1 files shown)


src/connections/sources/catalog/libraries/mobile/react-native/index.md
```diff
@@ -510,6 +510,8 @@ To collect the Android Advertising ID provided by Play Services, Segment provide
 No, only the plugins listed above are supported in device-mode for Analytics React Native 2.0.
 ### Will I still see device-mode integrations listed as `false` in the integrations object?
 When you successfully package a plugin in device-mode, you won't see the integration listed as `false` in the integrations object for a Segment event. This logic is packaged in the event metadata, and isn't surfaced in the Segment debugger.
+### Why are my IDs not set in UUID format?
+Due to [issues](https://github.com/segmentio/analytics-react-native/blob/master/packages/core/src/uuid.ts#L5){:target="_blank"} with the React Native bridge, Segment doesn't use UUID format for `anonymousId`s and `messageId`s in local development. These IDs will be set in UUID format for your live app.
```
stayseesong, line 514:
```suggestion
Due to [limitations](https://github.com/segmentio/analytics-react-native/blob/master/packages/core/src/uuid.ts#L5){:target="_blank"} with the React Native bridge, Segment doesn't use UUID format for `anonymousId`s and `messageId`s in local development. These IDs will be set in UUID format for your live app.  
```

stayseesong (approved):

PR #3539, opened 2022-09-22
Update Shopify Littledata 

edwardupton:
### Proposed changes

- updating links to Littledata help guides
- clarity around customer `state` property
- removing references to Shopify liquid files, which are no longer edited


markzegarelli:
Opened #3543 to make edits and deploy.

PR #3621, opened 2022-10-07
Scaffold Google Sheets documentation

kdaswani:
### Proposed changes
StratConn built the Google Sheets destination specifically for RETL. It is currently in Private Beta and being used by some RETL pilot customers. This is our first draft of the public docs so that the RETL team has something easy to share with customers. We want to keep these hidden/private but accessible via URL. I believe I've set the correct metadata to keep these hidden but let me know if I missed something!

Since this is in Private Beta, is there a way to ensure the action-fields components table is rendered? Do we need to run something extra to pull that autogenerated table in? Thanks!

### Merge timing
- ASAP once approved


markzegarelli (approved):

markzegarelli:
@kdaswani I'll merge this now for deploy on Tuesday. I've also added the private metadata so the available Action shows up as it should.

PR #3662, opened 2022-10-14
Add example that shows how to set the host to the EU endpoint

Gavin-Fitzpatrick:
We're facing a lot of confusion from customers on how to change the host to point towards to EU. Usually when we show them a code example they understand.

### Proposed changes

I've added an example of how to set the host to the EU endpoint in the Regional Section to make it quicker and easier for customers to implement. Changing the host changes from library to library with no way for the customer know other than to check our integration code or testing multiple versions of the URL.

### Merge timing
ASAP once approved


src/connections/sources/catalog/libraries/server/node/index.md
```diff
@@ -36,6 +36,14 @@ This will create an instance of `Analytics` that you can use to send data to Seg
 For Business plans with access to [Regional Segment](/docs/guides/regional-segment), you can use the `host` configuration parameter to send data to the desired region:
 1. Oregon (Default) — `api.segment.io/v1`
 2. Dublin — `events.eu1.segmentapis.com/v1/`
```
markzegarelli, line 38:
```suggestion
2. Dublin — `events.eu1.segmentapis.com`
```

markzegarelli (approved):

PR #3682, opened 2022-10-19
Add a note that explains why Segment does not map Identify to LaunchDarkly (Actions)

markzegarelli:

### DIFF
--- a/src/connections/destinations/catalog/actions-launchdarkly/index.md
+++ b/src/connections/destinations/catalog/actions-launchdarkly/index.md
@@ -1,7 +1,6 @@
 ---
 title: LaunchDarkly (Actions) Destination
 hide-boilerplate: true
-hide-dossier: true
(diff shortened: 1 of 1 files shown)


src/connections/destinations/catalog/actions-launchdarkly/index.md
```diff
@@ -43,11 +42,13 @@ To get started with LaunchDarkly (Actions):
 
 {% include components/actions-fields.html %}
 
+> info "A note about Identify calls"
+> LaunchDarkly does not store user attributes for use with flag evaluations. As a result, Segment does not support mapping Identify calls to LaunchDarkly actions.
```
stayseesong, line 46:
```suggestion
> LaunchDarkly doesn't store user attributes for use with flag evaluations. As a result, Segment doesn't support mapping Identify calls to LaunchDarkly actions.
```

stayseesong (approved):

PR #3753, opened 2022-11-01
Describe Segment -> Facebook user data mapping

markzegarelli:


kdaswani:
I think we should avoid adding this in @markzegarelli since customers can tweak any of the default mappings with actions. Was there a JIRA ticket requesting this update?

PR #3777, opened 2022-11-07
add logrocket documentation

nminchow:
### Proposed changes

Added docs for the LogRocket destination.

### Merge timing
asap

### Related issues (optional)

https://github.com/segmentio/action-destinations/pull/835


markzegarelli:
Thanks @nminchow! I've opened #3778 so that we can push updates. We'll get this deployed tomorrow, but it will remain hidden from search and nav until it enters Public or Public Beta status.

PR #3819, opened 2022-11-15
Add docs for new akita user tracking destination

tcgilbert:
### Proposed changes

Add docs for new akita user tracking destination

### Merge timing
asap


src/connections/destinations/catalog/akita-user-tracking/index.md
```diff
@@ -0,0 +1,84 @@
...
+```
+
+Segment sends Group calls to Akita as an `account` event.
+
+You may find a list of useful information to send as traits in your Group call in the [Segment > Configure](https://beta.akitaapp.com/segment){:target="_blank"} tab in Akita.
+
+Accounts described using the Group event are added to Akita as Accounts.
+
+> info ""
+> You must send Group events to Segment.com for the Segment <> Akita integration to work correctly.
```
markzegarelli, line 61: @tcgilbert This doesn't make sense as written. Where do Group events / calls need to be sent? Akita?

tcgilbert: Yeah, this would be better: 

"Groups identified through Segment group events are added to Akita as Accounts"

markzegarelli (approved):

PR #3911, opened 2022-12-06
Adding docs for Segment RETL destination

saitwilio:
### Proposed changes

Segment (Actions) are in private beta. We are moving to public beta on December 13.

Since this destination is in Private Beta, can we make sure that the dossier and other tables (other versions of destination, settings, actions, fields) get pre-populated upon deploy just in case the partner portal status hasn't updated in time? Thank you!

### Merge timing
- On a specific date - This should go out on the Dec 13 deploy as that is when we move the Segment destination to Public Beta.

### Related issues (optional)
new-integration

### DIFF
--- a/src/connections/destinations/catalog/actions-segment/index.md
+++ b/src/connections/destinations/catalog/actions-segment/index.md
@@ -3,6 +3,7 @@ title: Segment Destination
 hide-boilerplate: true
 hide-dossier: false
 id: 6371eee1ae5e324869aa8b1b
+private: true
 ---
 
 The Segment destination enables you to mold data extracted from your warehouse into [Segment Spec](/docs/connections/spec/) API calls that can be processed by Segment's Tracking API.


kdaswani:
Going to review tomorrow Thurs 12/8. Let's hold off on merging/deploying this until next Tues 12/13 please. Thanks!

kdaswani (approved):

src/connections/destinations/catalog/actions-segment/index.md
```diff
@@ -33,4 +33,4 @@ The Segment destination enables you to mold data extracted from your warehouse i
 ## FAQ & Troubleshooting
 
 ### API Calls and MTUs
```
James9446, line 35: I believe the plan is to only charge MTUs if the customer connects the source to a "classic" destination. If they only connect it to Engage then I don't think there would be any potential additional MTUs

src/connections/destinations/catalog/actions-segment/index.md
```diff
@@ -5,24 +5,32 @@ hide-dossier: false
...
+2. Click **Add Destination** in top-right corner.
+3. Select the Segment destination, click **Next**, and select the warehouse source that will send data to the Segment destination. If you have not set up a warehouse source, follow the steps in the Reverse ETL documentation on [Getting started](/docs/reverse-etl/#getting-started).
+4. On the **Settings** tab, name your destination, input the Write Key from the source created above, select an endpoint region, and click **Save Changes**.
+5. On the **Mappings** tab, click **Add Mapping**. Select a data model and the API call type you want to map. You can configure multiple mappings.
+6. Enable the destination and configured mappings.
 
-{% include components/actions-fields.html %}
+{% include components/actions-fields.html settings="true"%}
 
 ## FAQ & Troubleshooting
```
James9446, line 33: It might be worth calling out that the "Send test record" feature sends **live events** to the source. So for example, let's say a customer sets everything up and has a "classic" destination configured. Then at a later time they add a new mapping . If they use the "Send test record" feature, that test will be sent to HTTP API source and subsequently sent downstream to any connected destinations.

src/connections/destinations/catalog/actions-segment/index.md
```diff
@@ -5,24 +5,32 @@ hide-dossier: false
...
+2. Search for **HTTP API** in the Sources Catalog, select the HTTP API source, and click **Add Source**.
+3. Follow the steps to name your source. This is the source that will receive API calls from the Segment destination.
+4. Copy the Write Key on the **Overview** tab. You will need this when you set up the Segment destination.
+
+### Connect the Segment destination
+1. From the Segment web app, navigate to **Reverse ETL > Destinations**.
+2. Click **Add Destination** in top-right corner.
+3. Select the Segment destination, click **Next**, and select the warehouse source that will send data to the Segment destination. If you have not set up a warehouse source, follow the steps in the Reverse ETL documentation on [Getting started](/docs/reverse-etl/#getting-started).
+4. On the **Settings** tab, name your destination, input the Write Key from the source created above, select an endpoint region, and click **Save Changes**.
+5. On the **Mappings** tab, click **Add Mapping**. Select a data model and the API call type you want to map. You can configure multiple mappings.
```
James9446, line 28: If we assume that the customer is connecting this to a new model, then at the moment (though this may change) once you enable everything, the initial sync will kickoff. After that, only net new records will be sent (as a result of subsequent syncs). Therefore, it might be best to recommend all of the desired mappings be configured _before_ the destination is enabled. 

Here is an example of how this issue could play out:
1. Create a new model
2. Create a mapping for a given method (e.g. identify)
3. Enable that mapping
4. Data syncs for the first time and reaches the debugger
5. Then you create a new mapping for a different method (e.g. track)
6. Enable that mapping
7. No data will be sent to the debugger unless a new record is Added or Updated (because the new model has already synced)

kdaswani: @James9446 sounds good. Can you make the suggested edit to reflect this?

src/connections/destinations/catalog/actions-segment/index.md
kdaswani: This is a good point, but it's true of all Actions destinations/testers so not sure we really need to call it out one-off in this doc.

kdaswani: Maybe it's worth putting a note about that in these docs instead, if you want to file a separate PR for that: https://segment.com/docs/connections/destinations/actions/#customize-mappings

src/connections/destinations/catalog/actions-segment/index.md
kdaswani: This is only once we build a separate Segment Profiles destination. Today, there's no way for the team to bypass additional API calls/MTUs when connected to Engage since the data still needs to pass through the front door (i.e. HTTP API source). This is why our team is building a separate Segment Profiles destination in Q1 that would ideally allow connecting to only Engage spaces and bypass additional MTU/API calls costs.

markzegarelli (approved):

kdaswani:
Looks all set to me - thank you for your help James!

stayseesong (approved):

PR #3943, opened 2022-12-13
Clarified on GA4 sessions tracking implementation

050phy:
### Proposed changes

Dividing the implementation information into 2 sub-headings:  
1) tracking purely server side and it's limitations (not all user dimensions are available)  
2) using gtag.js and GA4 destination (client & server-side).

### Merge timing

No rush!

### Related issues (optional)

Previous version implied that a pure server side implementation was not possible and you could only get sessions tracking/reports with gtag.js on the client side. 

A Slack thread with Kiara (https://twilio.slack.com/archives/CQ1F92KUG/p1670813658588399) clarified that it is possible to do a pure server side implementation of GA4 with the caveat that the customer will not have all sessions data available in reports. The hope is that the proposed changes will make the differentiation clearer.


kdaswani:
This is awesome - very well written @050phy. Thank you! Looks good from my end. Will approve and let the docs team review grammar/style as well. Thanks again!

kdaswani (approved):

stayseesong (approved):

PR #3968, opened 2022-12-19
Add FAQ to HubSpot destination doc

prigiattiperrut:
### Proposed changes

Added a new topic to the Troubleshooting section "I Don't See My Data In HubSpot" as some customers are not seeing data in HubSpot because they have enabled the EU feature in Segment, but their HubSpot account is not enabled for the EU feature. This is causing events to be dropped on Hubspot side.

### DIFF
--- a/src/connections/destinations/catalog/hubspot/index.md
+++ b/src/connections/destinations/catalog/hubspot/index.md
@@ -172,6 +172,11 @@ Segment can send the following group traits to HubSpot as special properties:
 
 ## Troubleshooting
 
+
+### I Don't See My Data In HubSpot
+
+If you aren’t seeing your data arrive in HubSpot, we recommend you start by taking a look at your HubSpot configuration in Segment, and if you have enabled the option "Enable HubSpot SDK for EU" please confirm with the HubSpot team first that your account is enabled for EU data residency. 
+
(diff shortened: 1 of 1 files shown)


src/connections/destinations/catalog/hubspot/index.md
```diff
@@ -172,6 +172,11 @@ Segment can send the following group traits to HubSpot as special properties:
 
 ## Troubleshooting
 
+
+### I Don't See My Data In HubSpot
+
+If you aren’t seeing your data arrive in HubSpot, we recommend you start by taking a look at your HubSpot configuration in Segment, and if you have enabled the option "Enable HubSpot SDK for EU" please confirm with the HubSpot team first that your account is enabled for EU data residency.
```
markzegarelli, line 178:
```suggestion
If you don't see your data in HubSpot, verify the HubSpot configuration in your Segment workspace. If you've selected the **Enable HubSpot SDK for EU** option, confirm with HubSpot that your HubSpot account is enabled for EU data residency.
```

Few updates:
- Active voice
- avoid "we"
- Instruct, don't ask (avoid "please")

markzegarelli (changes requested):

PR #4015, opened 2023-01-09
Describe why `root` violations may appear in Protocols

niallzato:
### Proposed changes
Add info around Root Violation

### DIFF
--- a/src/protocols/faq.md
+++ b/src/protocols/faq.md
@@ -85,7 +85,7 @@ Must validate all the schemas
 (root)
 Must validate "then" as "if" was valid
 ```
-These will be related to your common JSON Schema if you have applied custom rules. In this instance (root), refers to the top level of the JSON object (Segment event). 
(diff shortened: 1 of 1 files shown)


src/protocols/faq.md
```diff
@@ -76,6 +76,17 @@ The Protocols Violations page shows a live count for violations. You can adjust
...
+### Why do I see root listed on my Violations page?
+You may see violations related to (root). For example:
+```js
+(root)
+Must validate all the schemas
+// Or
+(root)
+Must validate "then" as "if" was valid
+```
+These will be related to your common JSON Schema if you have applied custom rules. In this instance (root), refers to the top level of the JSON object (Segment event).
```
markzegarelli, line 88:
```suggestion
These violations are related to your common JSON Schema if you've applied custom rules. In this instance (root), refers to the top level of the JSON object (Segment event). 
```

markzegarelli (changes requested):

PR #4018, opened 2023-01-09
What to do when credentials have updated/changed

daldeano-segment:
### Proposed changes

Updating Getting Started Section with the following Information: 

**Note:**  It's important to note that when a connected account changes or updates their credentials, a new Facebook Lead Ads source will have to be created instead of solely reauthorizing the account as this is how FB has decided to operate.  Reauthorization implies that a new token gets generated, but the token is not the sole link between Segment and Facebook. A Segment app is installed onto an FB Account and that will be the sprouting point from where tokens are generated.

The reason we can't disconnect and reconnect the CRM on Facebook with reauthorization is because the creation of a new token will mean to invalidate the rest of the tokens attached to the same Facebook account. For example, let's imagine you have a Facebook Account with three accounts: Account A, Account B, and Account C. You connect Segment to all of them and then generate tokens respectively. Later on, you decide for whatever reason to reauthorize the "Account A" page, the other tokens will be invalidated, and the pages will stop working.  

Customer


charholt (approved):
looks great!

markzegarelli:
Hi @daldeano-segment, thanks for the PR!

In future PRs though, please wait for approval from a Docs team member before you merge. There are a few fixes we need to add here, that our grammar linter caught.

I will fix those in a separate PR, but you can see the warnings on the Files Changed tab.

PR #4020, opened 2023-01-10
Replace public-segment-devcenter-production S3 URLs with CloudFront URLs [DC-265]

konoufo:
This PR replaces the hardcoded S3 URLs pointing to bucket `public-segment-devcenter-production` with corresponding CloudFront URLs.

[_Created by Sourcegraph batch change `mohamed.coulibali/replace-s3-with-cloudfront`._](https://sourcegraph.segment.com/users/mohamed.coulibali/batch-changes/replace-s3-with-cloudfront)

### DIFF
--- a/templates/sources.example.yml
+++ b/templates/sources.example.yml
@@ -238,8 +238,8 @@ sources:
   categories:
   - Email Marketing
   logos:
-    logo: https://public-segment-devcenter-production.s3.amazonaws.com/485e07a1-bb4d-4718-97f5-4ffffc46387d.svg
-    mark: https://public-segment-devcenter-production.s3.amazonaws.com/480267ec-792a-43d2-a341-c73316b7b0db.svg
+    logo: https://cdn-devcenter.segment.com/485e07a1-bb4d-4718-97f5-4ffffc46387d.svg
+    mark: https://cdn-devcenter.segment.com/480267ec-792a-43d2-a341-c73316b7b0db.svg
 - name: catalog/sources/looker
   display_name: Looker
   description: This source is in beta! Looker is a business intelligence software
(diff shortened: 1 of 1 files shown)


markzegarelli:
@konoufo These files are generated automatically with output from the Public API. Are these URLs updated on the backend / in Partner Portal?

konoufo:
> @konoufo These files are generated automatically with output from the Public API. Are these URLs updated on the backend / in Partner Portal?

@markzegarelli Hum, interesting! Thank you for pointing this out. We are planning to replace all control-plane-db entries containing S3 URLs related to this bucket next week. Which should probably do it. But for awareness, would you guide me to where exactly these are generated ?

markzegarelli:
@konoufo Cool! Updating them directly in control-plane should do it, as I believe that is what feeds the Public API. I'm not sure where these are generated though. I know the Dev Portal supports uploading image files for logos, but I do not know how that ties into the back end.

Docs are very much end-of-chain in this process.

konoufo:
@markzegarelli ok I see there is a script [here](https://github.com/segmentio/segment-docs/blob/develop/scripts/catalog_papi.js) that generates these files. Is the script run manually or at deploy-time on Netlify ?
In either case then can you accept this PR so that I will just make sure to remember to merge these changes here when we are done with the migration on control-plane-db. Worst case scenario it will trigger the necessary re-deploy.
I want to make sure we don't forget any broken links when we end up blocking direct S3 URLs.

markzegarelli:
@konoufo We run the script manually before we deploy, so that updates are committed to the repository.

My preference for this PR is to close it, since once the updates are done in control-plane, we will get the updates that way. We've had issues in the past when we manually update this file.

konoufo:
> @konoufo We run the script manually before we deploy, so that updates are committed to the repository.
> 
> My preference for this PR is to close it, since once the updates are done in control-plane, we will get the updates that way. We've had issues in the past when we manually update this file.

Closing this as per @markzegarelli recommendation.

konoufo:
@markzegarelli The autogenerated have been updated as you said they would, thank you. I'm reopening after noticing that at least `examples/sources.example.yml` was not autogenerated. I'm dropping changes on all other files but this one.

markzegarelli (approved):

PR #4021, opened 2023-01-10
Add information about deleting/changing ExternalID's

olesia-ag:
### Proposed changes

Added FAQ with information about deleting/changing ExternalID's


olesia-ag:
the change will go to another docs page - https://github.com/segmentio/segment-docs/pull/4034

PR #4030, opened 2023-01-11
Add guidance for the number of sources a user should create

sarahrudy:
Add paragraph to sources page about whether to have one source or multiple sources

### Proposed changes

I've seen several tickets where customers ask if they should set up one source or multiple sources, so it wouldn't hurt to have this somewhere in the docs. Would it make more sense to add a FAQs section to the bottom of the page? 

### Merge timing
- ASAP once approved

### Related issues (optional)
N/A


forstisabella:
Hi @sarahrudy: this looks similar to #4053: should this information be included in the source docs twice?

sarahrudy:
Hi @forstisabella, thank you for pointing this out! I'm not sure how both of these got created. Since the other one was merged, this one can be closed!

PR #4031, opened 2023-01-11
add details about data type mismatch

muhammadatifjav:
### Proposed changes

Warehouse data type mismatch is a common issue that the customers face. We have macros/internal docs to explain the issue, and offer possible resolution, but it can be helpful if we can explain why this happens and quicker to resolve if the customer uses the template when reaching out to resolve this data type mismatch.

The changes proposed include:
- Structuring "New Columns", "Supported Data Types", and "Column Sizing" under "Schema Evolution and Compatibility" heading
- Adding content to explain how data type mismatches occurs, steps to resolve and considerations to take in account, and also give a format to the customer to use when reaching out with requests

### DIFF
--- a/src/connections/storage/warehouses/schema.md
+++ b/src/connections/storage/warehouses/schema.md
@@ -427,7 +427,7 @@ Data types are set up in your warehouse based on the first value that comes in f
 
 In cases where a data type is determined incorrectly, the support team can help you update the data type. As an example, if a field can include float values as well as integers, but the first value we received was an integer, we will set the data type of the field to integer, resulting in a loss of precision. 
 
-To update the data type, the support team will update the internal schema that Segment uses to infer your warehouse schema. We will start syncing the data with the correct data type after the change is made. However, if you want to backfill all historical data correctly, it will be required to drop the impacted tables on your end so Segment can recreate them in the correct datatype, and then backfill those tables. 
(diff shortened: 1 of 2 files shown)


charholt (approved):

src/connections/storage/warehouses/schema.md
```diff
@@ -423,10 +423,16 @@ The data types that Segment currently supports include:
 
 #### `varchar`
 
-> note " "
-> To change data types after they've been determined, please reach out to [Segment Support](https://segment.com/help/contact) for assistance. 
+Data types are set up in your warehouse based on the first value that comes in from a source. For example, if the first value that came in from a source was a string, Segment would set the data type in the warehouse to `string`. 
+
+In cases where a data type is determined incorrectly, the support team can help you update the data type. As an example, if a field can include float values as well as integers, but the first value we received was an integer, we will set the data type of the field to integer, resulting in a loss of precision. 
+
+To update the data type, the support team will update the internal schema that Segment uses to infer your warehouse schema. We will start syncing the data with the correct data type after the change is made. However, if you want to backfill all historical data correctly, it will be required to drop the impacted tables on your end so Segment can recreate them in the correct datatype, and then backfill those tables.
```
forstisabella, line 430: Hi @muhammadatifjav, I had a quick question about this part - after a customer drops the impacted tables, would Segment automatically recreate the tables and backfill them, or would a customer need to reach out to facilitate the backfill?

muhammadatifjav: After customer drops the impacted table, Segment will automatically recreate the table and start syncing new data automatically. However, if they want historical data to be corrected, then they will need to drop the table, and then confirm it to us and we will then run the backfill. I've made another commit to make it more clearer, lmk if that is better

forstisabella (approved):

PR #4046, opened 2023-01-12
Add section to Troubleshooting for solving stuck syncs to Facebook

bobbyatsegment:
### Proposed changes

Customers frequently write in to Segment Support on issues related to audiences failing to sync to Facebook Custom Audiences, and are perplexed by the "Internal Error" that surfaces for syncs that fail for longer than 14 days after a series of successful syncs, for example:

image

Adding a solution to this issue along with some brief context as to why it occurs could help to serve customers better in handling this issue without having to contact support, should they choose to explore the first solution of deleting and re-creating the audience.


cmastr:
@bobbyatsegment does this works till need to be done? This PR is quite old and still in **draft** state. We plan to close this PR on Friday January 26, 2024. If the work still needs to be done, please let us know.

PR #4052, opened 2023-01-13
Note that API Keys are region specific in Amplitude

sarahrudy:
Add info block to point out that API key is region specific

### Proposed changes

Add info block to point out that API key is region specific

### Merge timing
- ASAP once approved


sarahrudy:
It would be great to add this directly under the Destination Settings block, but I'm not sure how to add it since that block is coming from {%capture log-event-details %}

markzegarelli:
Hi @sarahrudy!

Would it be possible to update the description of the API Key value in Partner Portal to include this information? It might make more sense contextually there.

sarahrudy:
Great idea, Mark! @tcgilbert said he was able to make that change directly in Partner Portal.

PR #4053, opened 2023-01-13
Add guidance around the number of sources a user should create

sarahrudy:
Add info block about whether customer should set up one source or multiple sources

### Proposed changes

Add info block about whether customer should set up one source or multiple sources

### Merge timing
- ASAP once approved?

### DIFF
--- a/src/connections/sources/index.md
+++ b/src/connections/sources/index.md
@@ -71,6 +71,13 @@ To create a source:
 4. Enter a name for your source as well as any information on the setup page.
 5. Click **Add Source**.
 
+> info ""
+> **One source or multiple sources?** 
+> Ultimately, it's up to you, but it really depends on how you want to organize your data. We suggest that you create one source for each type of data you want to collect. For example, a customer might have one source for all of their website tracking and a different source for any mobile tracking. We typically recommend having one source for each type of data (different libraries) for the following reasons:
+> - Debugger ease of use - mixing libraries/sources on a single API key means you’re heavily reliant on filtering to actually test events
+> - Flexibility sending data to different projects - if you want to have different warehouse schemas, analytics projects, etc, having multiple sources would create this separation
+> - More control - as your account grows with the number of destinations you enable, having separate sources allows you to have more control
+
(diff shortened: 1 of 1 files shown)


src/connections/sources/index.md
```diff
@@ -71,6 +71,13 @@ To create a source:
 4. Enter a name for your source as well as any information on the setup page.
 5. Click **Add Source**.
 
+> info ""
+> **One source or multiple sources?**
```
markzegarelli, line 75:
```suggestion
```

src/connections/sources/index.md
```diff
@@ -71,6 +71,13 @@ To create a source:
 4. Enter a name for your source as well as any information on the setup page.
 5. Click **Add Source**.
 
+> info ""
```
markzegarelli, line 74:
```suggestion
> info "One source or multiple sources?"
```

src/connections/sources/index.md
```diff
@@ -71,6 +71,13 @@ To create a source:
...
+> **One source or multiple sources?** 
+> Ultimately, it's up to you, but it really depends on how you want to organize your data. We suggest that you create one source for each type of data you want to collect. For example, a customer might have one source for all of their website tracking and a different source for any mobile tracking. We typically recommend having one source for each type of data (different libraries) for the following reasons:
```
markzegarelli, line 76:
```suggestion
> Segment suggests that you create one source for each type of data you want to collect. For example, a you might have one source for all of their website tracking and a different source for any mobile tracking. Creating one source per data type provides the following benefits:
```

markzegarelli: Important to remember that we're our customers and prospects are the main audience for our docs. We typically just use `you` as opposed to `a customer`.

markzegarelli (changes requested):

src/connections/sources/index.md
sarahrudy: Thank you for pointing that out!

markzegarelli (approved):

PR #4074, opened 2023-01-18
Update faq.md

bcaudillo:
### Proposed changes

I added a section into Protocols FAQ to speak to what happens if Protocols is disabled and then re-enabled. If this should be internal, please let me know.


cmastr:
@bcaudillo does this works till need to be done? This PR is quite old and still in draft state. We plan to close this PR on Friday January 26, 2024. If the work still needs to be done, please let us know.

PR #4078, opened 2023-01-18
Add info note specifying supported data types for mapping operators

bobbyatsegment:
### Proposed changes

Several customers have written in to support expressing confusion on why certain operators in event triggers don't behave as expected in regard to matching on data types. For example, a trigger with an `is` operator will not match on a numeric data type, only a string:

Matching on string:
match_on_str

Not matching on number:
no_match_num

The expectations for these mappings aren't directly mentioned in the docs - adding a section that highlights operators and their allowed data types should help to clarify expected behavior. 

Slack thread for context: https://twilio.slack.com/archives/CC97A542H/p1674058095447979


spencerattick (approved):

markzegarelli:
@bobbyatsegment This is great info! I reformatted things a bit to improve readability. The nested list in the info box didn't render that great, so I moved it into a table.
Screen Shot 2023-01-19 at 9 35 33 AM

markzegarelli (approved):

PR #4097, opened 2023-01-20
update api links

tcgilbert:
### Proposed changes

Looks like there was a site wide change to update any mention of the "Config API" with the "Public API" back in september. Unfortunately, the tokens generated from the "enable with oauth" flow are only compatible with the config api, so this has caused some confusion for partners with 1.0 integrations. 

### Merge timing
asap

### DIFF
--- a/src/partners/enable-with-segment.md
+++ b/src/partners/enable-with-segment.md
@@ -2,7 +2,7 @@
 title: Enable with OAuth
 ---
 
-Enable with OAuth, supported with the [Public API](/docs/api/public-api), allows partners to build seamless flows for customers to implement and enable integrations within their workspace. This is a [post-launch requirement](/docs/partners/#post-launch) for a partner to graduate from public beta to public.
+Enable with OAuth, supported with the [Config API](/docs/api/config-api), allows partners to build seamless flows for customers to implement and enable integrations within their workspace. This is a [post-launch requirement](/docs/partners/#post-launch) for a partner to graduate from public beta to public.
 
 ## Concepts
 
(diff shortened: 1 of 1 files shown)


src/partners/enable-with-segment.md
```diff
@@ -2,25 +2,25 @@
 title: Enable with OAuth
 ---
 
-Enable with OAuth, supported with the [Public API](/docs/api/public-api), allows partners to build seamless flows for customers to implement and enable integrations within their workspace. This is a [post-launch requirement](/docs/partners/#post-launch) for a partner to graduate from public beta to public.
+Enable with OAuth, supported with the [Config API](/docs/api/config-api), allows partners to build seamless flows for customers to implement and enable integrations within their workspace. This is a [post-launch requirement](/docs/partners/#post-launch) for a partner to graduate from public beta to public.
```
rchinn-segment, line 5: Hey @tcgilbert, just want to confirm that graduating from public beta to public is correct, as opposed to something like private beta to public.

tcgilbert: "public beta to public" is the standard we set for dev center 1.0 destinations. It is slightly different with 2.0 integrations, in that there is no `enable with oauth` requirement.

Being that 1.0 is legacy I'd be okay with removing the last sentence entirely: 

-  "This is a [post-launch requirement](/docs/partners/#post-launch) for a partner to graduate from public beta to public."

rchinn-segment: Thanks for clarifying. I removed the last sentence and will include these updates in our next deploy.

rchinn-segment (approved):

PR #4102, opened 2023-01-23
Update AJS FAQs to include adding new context fields

ashton-huxtable:
### Proposed changes

Added an FAQ that speaks directly to adding new context, rather than overwriting existing context. Many customers don't understand that they can do both. 

### Merge timing
- ASAP once approved

### Related issues (optional)

n/a

### DIFF
--- a/src/connections/sources/catalog/libraries/website/javascript/troubleshooting.md
+++ b/src/connections/sources/catalog/libraries/website/javascript/troubleshooting.md
@@ -131,6 +131,22 @@ analytics.track("Receipt Viewed", {}, {
 
 This works for any [context field](/docs/connections/spec/common/#context) that Segment automatically collects.
 
+### Can I add context fields that do not already exist?
+
+Yes. Similar to overwriting context, you can add context fields by passing them into the options object as the third argument of the event call. For example, the analytics.js library does not automatically collect location information. So, if you wanted to add this into your context object, you would pass it into the third argument as in the example below:
+
+```js
+analytics.track("Order Completed", {}, {
+    location: {
+        latitude: '39.7392',
+        longitude: '104.9903'
+    }
+})
+```
+
+It is important to note that if you are not passing in properties, you still need to pass an empty object in it's place as shown in the above example. 
+
+
 ## Known issues:
 
 [Review and contribute to these on GitHub](https://github.com/segmentio/analytics.js/issues)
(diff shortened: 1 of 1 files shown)


src/connections/sources/catalog/libraries/website/javascript/troubleshooting.md
```diff
@@ -131,6 +131,24 @@ analytics.track("Receipt Viewed", {}, {
 
 This works for any [context field](/docs/connections/spec/common/#context) that Segment automatically collects.
 
+### Can I add context fields that do not already exist?
+
+Yes. Similar to overwriting context, you can add context fields by passing them into the options object as the third argument of the event call. For example, the analytics.js library does not automatically collect location information. So, if you wanted to add this into your context object, you would pass it into the third argument as in the example below:
```
markzegarelli, line 136:
```suggestion
Yes. Similar to overwriting context, you can add context fields by passing them into the options object as the third argument of the event call. For example, the analytics.js library does not automatically collect location information. To add this into the context object, pass it into the third argument as in the example below:
```

src/connections/sources/catalog/libraries/website/javascript/troubleshooting.md
```diff
@@ -131,6 +131,24 @@ analytics.track("Receipt Viewed", {}, {
...
+```js
+analytics.track("Order Completed", {}, {
+    location: {
+        latitude: '39.7392',
+        longitude: '104.9903'
+    }
+})
+```
+
+It is important to note that if you are not passing in properties, you still need to pass an empty object in it's place as shown in the above example.
```
markzegarelli, line 147:
```suggestion
> info ""
> You must pass the context object with the call, event if it's empty, as shown by the empty object in the example above.
```

src/connections/sources/catalog/libraries/website/javascript/troubleshooting.md
```diff
@@ -131,6 +131,24 @@ analytics.track("Receipt Viewed", {}, {
...
+    location: {
+        latitude: '39.7392',
+        longitude: '104.9903'
+    }
+})
+```
+
+It is important to note that if you are not passing in properties, you still need to pass an empty object in it's place as shown in the above example. 
+
+Additionally, custom context fields may not be forwarded to all destinations, as some destinations accept properties only.
```
markzegarelli, line 149:
```suggestion
Some destinations accept properties only. As a result, custom context fields you add may not forward to these destinations.
```

markzegarelli (changes requested):

markzegarelli (approved):

PR #4103, opened 2023-01-23
Update faqs.md

Hopes-Chantel:
### Proposed changes

I updated Engage FAQ to include information on how to delete unwanted traits from UI. I have received multiple tickets where customers have asked how this could be achieved and I think since it is a simple process it should be available via the public docs. (Please let me know if you think there is a better place for this information)


cmastr:
image

PR #4106, opened 2023-01-23
Added new NAT CIDR block to allowlist instructions

etiennep:
### Proposed changes

Added NAT IP address CIDR blocks to allowlist instructions. A newer larger IP address block will replace the existing single IP address in the near future to provide greater resiliency and scalabiliy to our networking infrastructure.

Customers will need to add the new CIDR range to their inbound rules before we make the switch.

### Merge timing

- ASAP once approved. The sooner this gets updated, the less disruptive it will be for new customers.


src/connections/storage/catalog/redshift/index.md
```diff
@@ -86,7 +86,7 @@ Redshift clusters are created in a VPC subnet. To configure:
 
 4. Click the Security group in the list to access its settings.
 
-5. On the Inbound tab, add or edit a rule to enable Segment to write to your Redshift port from `52.25.130.38/32`. ![inbound](images/redshift05.png)
+5. On the Inbound tab, add rules to enable Segment to write to your Redshift port from `34.223.203.0/28` and from `52.25.130.38/32`. ![inbound](images/redshift05.png)
```
forstisabella, line 89:
```suggestion
5. On the Inbound tab, add rules to enable Segment to write to your Redshift port from `34.223.203.0/28` and `52.25.130.38/32`. ![inbound](images/redshift05.png)
```

forstisabella (approved):

PR #4108, opened 2023-01-23
Add details on Replay when connecting Engage Sources

lizkane222:
Adding the replay details in this section provides additional context around what occurs when a Source is added. The common misconception is that all of that source's data will be available in the Engage Space, but having this section added would subvert that misconception and enable customers to reach out preemptively with replay requests for historical data beyond 30 days.

### Proposed changes

Adding the replay details in this section provides additional context around what occurs when a Source is added to an Engage Space. The common misconception is that all of that source's data will be available in the Engage Space, but having this section added would subvert that misconception and enable customers to reach out preemptively with replay requests for historical data beyond 30 days.

Though some of this data can be found on the link provided, [Connect production sources](https://segment.com/docs/profiles/quickstart/#step-3-connect-production-sources), some key details in the setup might be overlooked in this possibly too simple of a setup.

### Merge timing
- ASAP once approved


cmastr:
@lizkane222  does this works till need to be done? This PR is quite old and still in draft state. We plan to close this PR on Friday January 26, 2024. If the work still needs to be done, please let us know.

cmastr (approved):

PR #4113, opened 2023-01-24
Fixed typos

sneha-shashidhar:


stayseesong (commented):
Hi @sneha-shashidhar, these are not typos. What's currently written is correct as it's in active format. What you have written is passive voice and we don't use passive voice for our docs

PR #4135, opened 2023-01-26
PlayerZero Web Docs

AnimeshKoratana:
### Proposed changes
We added a documentation page to the segment docs to complete our Segment Integration and prepare for it to go live. This enables Segment users to discover the PlayerZero + Segment integration and quickly enable it for themselves.

### Merge timing
- ASAP once approved: The integration is already live as a private beta link and has been used by multiple customers.

### DIFF
--- a/src/connections/destinations/catalog/actions-playerzero-web/index.md
+++ b/src/connections/destinations/catalog/actions-playerzero-web/index.md
@@ -1,0 +1,35 @@
+---
+title: PlayerZero Web (Actions)
+---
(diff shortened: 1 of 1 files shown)


markzegarelli:
Thanks @AnimeshKoratana! I've created #4136 so that we can push updates to the PR. I've gone ahead and added our Actions framework tooling to the doc, so that we can pull and display the destination's settings, presets, and action details. We will deploy it on Tuesday, January 31st.

PR #4158, opened 2023-01-31
Add warning about no longer being able to create Config API Tokens

bobbyatsegment:
### Proposed changes

Customers have written in to Segment Support asking how to create a Config API token, but since September 8, 2022, new Config API token generation has ceased - relevant Slack: https://twilio.slack.com/archives/C01BJ2ZCBHC/p1662576926040749 

Adding a warning to the Config API docs that directly mentions new Config API token generation has ceased (but that previously generated tokens still work) should help to alleviate customer confusion on this.

### DIFF
--- a/src/api/config-api/index.md
+++ b/src/api/config-api/index.md
@@ -6,6 +6,9 @@ redirect_from:
 
 {% include content/papi-ga.html %}
 
+> warning "Config API Tokens can no longer be generated"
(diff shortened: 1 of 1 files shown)


cmastr:
@bobbyatsegment does this works till need to be done? This PR is quite old and still in draft state. We plan to close this PR on Friday January 26, 2024. If the work still needs to be done, please let us know.

PR #4164, opened 2023-01-31
Correcet folder name and source name

tcgilbert:
### Proposed changes

Correct folder name and source name

### Merge timing
asap

### Related issues (optional)

One thing worth noting is that the Foursquare Pilgrim name is still in the search bar, hopefully this will fix that. 

Screen Shot 2023-01-31 at 4 37 13 PM


markzegarelli:
@tcgilbert Yes, I think this is the issue. I will also add a redirect from the `foursquare-pilgrim` url.

markzegarelli (approved):

PR #4195, opened 2023-02-08
Update actions.md

Panandhan22:
### Proposed changes

The mappings can't be triggered in certain sequence, even it is configured with the same event.

### DIFF
--- a/src/connections/destinations/actions.md
+++ b/src/connections/destinations/actions.md
@@ -66,6 +66,7 @@ To set up a new Actions-framework destination for the first time:
 
 > info ""
 > At least one mapping to handle a connected source's event(s) must be configured and enabled in an Actions-framework destination in order for data to be sent downstream.
+> The mappings can't be triggered in certain sequence even it is configured with the same event.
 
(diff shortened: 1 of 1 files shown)


src/connections/destinations/actions.md
```diff
@@ -66,6 +66,7 @@ To set up a new Actions-framework destination for the first time:
 
 > info ""
 > At least one mapping to handle a connected source's event(s) must be configured and enabled in an Actions-framework destination in order for data to be sent downstream.
+> Events will be sent downstream in the order in which they appear in the mappings UI. There is no mechanism through which you can control the order of events sent to the downstream destinations outside of that.
```
stayseesong, line 69:
```suggestion
> You must configure and enable at least one mapping to handle a connected source's event(s) in an Actions-framework destination in order for data to send downstream. 
> Events send downstream in the order in which they appear in the mappings UI. There is no mechanism through which you can control the order of events that send to the downstream destinations outside of that. 
```

stayseesong (changes requested):
changing to active voice

stayseesong (approved):

PR #4210, opened 2023-02-10
Copy a Tracking Plan into a Library

jfoskin:
### Proposed changes

I added a question to our Protocols FAQ. We get a number of tickets, asking if copying a Tracking Plan into a Library is possible.

### DIFF
--- a/src/protocols/faq.md
+++ b/src/protocols/faq.md
@@ -62,6 +62,10 @@ You can search in a Tracking Plan to find a specific event, and then copy the UR
 
 Yes. [Tracking Plan Libraries](/docs/protocols/tracking-plan/libraries/) makes it easy to create groups of events or properties that can be easily imported into multiple Tracking plans.
 
+### Can I copy a Tracking Plan into a library?
+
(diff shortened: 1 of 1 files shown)


src/protocols/faq.md
```diff
@@ -62,6 +62,10 @@ You can search in a Tracking Plan to find a specific event, and then copy the UR
 
 Yes. [Tracking Plan Libraries](/docs/protocols/tracking-plan/libraries/) makes it easy to create groups of events or properties that can be easily imported into multiple Tracking plans.
 
+### Can I copy a Tracking Plan into a library?
+
+No. Unfortunately it's not possible yet to automatically transfer events from a Tracking Plan to Libraries. Currently the best option for importing events into a new event library would be directly from a source.
```
forstisabella, line 67:
```suggestion
No. Unfortunately it's not yet possible to automatically transfer events from a Tracking Plan to Libraries. The best option for importing events into a new event library would be to import them directly from a source.
```

forstisabella (approved):

PR #4247, opened 2023-02-16
adding context to FAQ section for redshift serverless

charholt:
### Proposed changes

Added to FAQ that we do not support redshift serverless

### Merge timing
- ASAP once approved

### DIFF
--- a/src/connections/storage/catalog/redshift/index.md
+++ b/src/connections/storage/catalog/redshift/index.md
@@ -149,4 +149,4 @@ Segment does not currently support SSH tunneling to Redshift. You can usually al
 
 ### Do you support Redshift Serverless?
 
-Segment does not currently support Serverless Redshift. While you are able to set up the connection in the Segment UI, we do not have the functionality to query Redshift's SYS tables.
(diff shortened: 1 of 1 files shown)


src/connections/storage/catalog/redshift/index.md
```diff
@@ -146,3 +146,7 @@ You can also unload data to a s3 bucket and then load the data into another Reds
 ### Can I use an SSH tunnel to connect to my Redshift instance?
 
 Segment does not currently support SSH tunneling to Redshift. You can usually allow Segment's ETL to write to Redshift without leaving the cluster available to other connections by using IP level restrictions.
+
+### Do you support Redshift Serverless?
+
+Segment does not currently support Serverless Redshift. While you are able to set up the connection in the Segment UI, we do not have the functionality to query Redshift's SYS tables.
```
forstisabella, line 152: We try to avoid first person pronouns (for example 'we', 'I', 'us', 'ours') in the docs whenever possible, because it can create confusion!

```suggestion
Segment does not currently support Serverless Redshift. While you can set up the connection in the Segment app, Segment does not have the functionality to query Redshift's SYS tables.
```

PR #4265, opened 2023-02-21
Remove unsupported features

niallzato:
### Proposed changes

All the things removed I do not think applies to the plugin. Checking the code here: https://github.com/segment-integrations/analytics-swift-facebook-app-events/blob/main/Sources/SegmentFacebook/FacebookAppEventsDestination.swift

I do not see logic to transform events or the majority of the properties.


markzegarelli (approved):

alanjcharles (approved):
lgtm!

PR #4266, opened 2023-02-21
Update Swift Firebase

niallzato:
### Proposed changes
Checking the code here, I do not think we support transforming the traits: https://github.com/segment-integrations/analytics-swift-firebase/blob/main/Sources/SegmentFirebase/FirebaseDestination.swift#L52

Also removed references to Android behavior.


alanjcharles (approved):
lgtm!

PR #4267, opened 2023-02-21
Swift Mixpanel Updates

niallzato:
### Proposed changes

We do not seem to change the trait names, so removing that section.
No logic specifically for array.
Moved screen tracking to distinct section.

https://github.com/segment-integrations/analytics-swift-mixpanel/blob/main/Sources/SegmentMixpanel/MixpanelDestination.swift


alanjcharles (approved):
lgtm!

PR #4280, opened 2023-02-23
Update index.md

tanjinhong72:
### Proposed changes

In the past, there were some tickets on customer asking about sending null property and trait. Engineering has also confirmed that this is not possible.

Added information that sending of null property or trait is not possible.

In the following section:
> **Good to know**: For any of the different methods described on this page, you can replace the properties and traits in the code samples with variables that represent the data collected.

Added:
"Note that sending of null property or trait will not be possible as Guava’s immutable maps will reject the null value and GSON library use to serilaize the Java object will ignore the null value.  As a workaround, send an empty string instead of a null value in the property or trait."

### Related issues (optional)
(https://segment.zendesk.com/agent/tickets/498628)

### DIFF
--- a/src/connections/sources/catalog/libraries/server/java/index.md
+++ b/src/connections/sources/catalog/libraries/server/java/index.md
@@ -63,7 +63,7 @@ The following examples use [Guava's](https://github.com/google/guava) immutable 
 ## Identify
 
 > note ""
-> **Good to know**: For any of the different methods described on this page, you can replace the properties and traits in the code samples with variables that represent the data collected.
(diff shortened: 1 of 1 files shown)


stayseesong:
hi @tanjinhong72, is there a reason why you moved this note around and added it to the Regional Configuration section?

tanjinhong72:
Hi @stayseesong , sorry on the late reply. The 'Good to know' note is meant for identify and track types, so initially it feels better to have it above the Identify section. However, I do see the point that it looks even more unappropriate to be under Regional Configuration section. So it's fine to put the 'Good to know' note back under Identify section.

stayseesong (approved):

PR #4299, opened 2023-02-24
Changing a source type note

jfoskin:
A source type cannot be changed if you decide you would like to use another source in the future. You would have to create a new source if you would like to use a different source.

We get a number of tickets from customers asking if they can change their source type


src/connections/sources/index.md
```diff
@@ -76,6 +76,7 @@ To create a source:
 > - Debugger ease of use - mixing libraries/sources on a single API key means you’re heavily reliant on filtering to actually test events
 > - Flexibility sending data to different projects - if you want to have different warehouse schemas, analytics projects, etc, having multiple sources would create this separation
 > - More control - as your account grows with the number of destinations you enable, having separate sources allows you to have more control
+> - A source type cannot be changed if you decide you would like to use another source in the future. You would have to create a new source if you would like to use a different source.
```
forstisabella, line 79: Feel free to correct me if this isn't what you meant!

```suggestion
> - A source type cannot be changed once it is created. You must create a new source if you would like to use a different source type.
```

PR #4334, opened 2023-03-01
First party cookies

jfoskin:
Adding in information regarding additional cookies a customer may come across on their website.

We get tickets asking about additional cookies that are present on a customer's website, and they are not aware where these cookies are coming from


src/connections/sources/catalog/libraries/website/javascript/troubleshooting.md
```diff
@@ -157,6 +157,11 @@ analytics.track("Order Completed", {}, {
 
 Some destinations accept properties only. As a result, custom context fields you add may not forward to these destinations.
 
+### Why am I seeing additional cookies on my website?
+
+The AJS cookies being set under segment.com are first-party cookies, and they are part of our own implementation of Segment as well as the destination we are using. These cookies are not related to your implementation of Segment. You are only seeing them because you have visited Segment's domain using the same browser. They are being sent to the writekey connected to Segment's own workspace. They are associated with the events Segment tracks when you visit our domain, segment.com
```
rchinn-segment, line 162:
```suggestion
The AJS cookies being set under segment.com are first-party cookies. They are part of Segment's own implementation as well as the destination Segment uses. These cookies are not related to your implementation of Segment, and you only see them because you've visited Segment's domain using the same browser. They are sent to the writekey connected to Segment's own workspace, and are associated with the events Segment tracks when you visit segment.com.
```

rchinn-segment (approved):

PR #4376, opened 2023-03-08
Update Transformation doc

vanand17:
Update a line to provide example and clarify what we are explaining in our docs to avoid confusion.

### Merge timing
NA

### Related issues (optional)

NA


src/protocols/transform/index.md
```diff
@@ -95,7 +95,16 @@ Depending on the type of transformation you selected, you will need to enter the
...
+If you rename properties or traits within a selected event, click **+ Add Property**. The dropdown that appears contains the properties or traits sent with the selected event. Segment supports JSON Path notation to select nested objects up to four levels deep. For example, `order.id` selects the `id` property in the `order` object. Segment does not support `.$.` notation to select a property from an array of objects. For example, the following event will generate products.$.product_id 
+
+analytics.track('Example', {
+  products: [{
+      product_id: "123"
+  }],
+})
+```
+
+In this scenario, we do not support the transformation of product_id.
```
forstisabella, line 107:
```suggestion
```

src/protocols/transform/index.md
```diff
@@ -95,7 +95,16 @@ Depending on the type of transformation you selected, you will need to enter the
 After you select the scope, use the search box to choose the event to transform. You can **only** select a single track event, identify or group call. If you are renaming the event, simply enter the new name in the provided text field.
 
 * **Rename properties or traits:**
-If you rename properties or traits within a selected event, click **+ Add Property**. The dropdown that appears contains the properties or traits sent with the selected event. Segment supports JSON Path notation to select nested objects up to four levels deep. For example, `order.id` selects the `id` property in the `order` object. Segment does not support `.$.` notation to select a property from an array of objects.
+If you rename properties or traits within a selected event, click **+ Add Property**. The dropdown that appears contains the properties or traits sent with the selected event. Segment supports JSON Path notation to select nested objects up to four levels deep. For example, `order.id` selects the `id` property in the `order` object. Segment does not support `.$.` notation to select a property from an array of objects. For example, the following event will generate products.$.product_id
```
forstisabella, line 98:
```suggestion
If you rename properties or traits within a selected event, click **+ Add Property**. The dropdown that appears contains the properties or traits sent with the selected event. Segment supports JSON Path notation to select nested objects up to four levels deep. For example, `order.id` selects the `id` property in the `order` object. Segment does not support `.$.` notation to select a property from an array of objects. For example, the following event would be unable to generate products.$.product_id:
```

forstisabella (changes requested):
looks good to me pending a few small changes!

src/protocols/transform/index.md
vanand17: Hi @forstisabella Thanks for your feedback about the code block, will remember that for future changes.
Regarding your proposed line here, "For example, the following event would be unable to generate products.$.product_id"
That might not be in line with what we are saying. We are providing an example to let them know that is how a nested property would look in a payload, and that this is currently not supported in Transformations.
Hence, we are saying the following event will generate products.$.product_id , and after the code block, we mention that this product_id will not be supported in Transformations.

Let me know if I have misunderstood anything here.

forstisabella: Hi @vanand17 - if Segment does not currently support something, I might avoid including an example to minimize confusion. Reading "_For example, the following event will generate products.$.product_id:_" and then seeing immediately after the example that it's not supported is confusing!

You could try including info about what Segment **doesn't** support in a `warning` callout brings it to the user's attention and sets it apart from the surrounding content. 

Also - what happens if a user tries to use a nested Transformation?

vanand17: Hi @forstisabella I agree, however some customers are confused about the nested property, hence we thought providing an example would be a good way to clear that up. For example, they think product.product_id is the same as order.id
I hope that example explains the confusion we typically see with customers.
Users will not be able to select a nested property in transformation as it is not supported.

src/protocols/transform/index.md
```diff
@@ -95,7 +95,17 @@ Depending on the type of transformation you selected, you will need to enter the
...
```
forstisabella, line 98:
```suggestion
To rename properties or traits within a selected event, click **+ Add Property**. The dropdown that appears contains the properties or traits sent with the selected event. Segment supports JSON Path notation to select nested objects up to four levels deep. For example, `order.id` selects the `id` property in the `order` object. Segment does not support `.$.` notation to select a property from an array of objects. For example, the following event, which generates `products.$.product_id`, is unsupported.
```

forstisabella (commented):
Looks good to me pending one more change.

vanand17:
Looks good to me, thank you!

forstisabella (approved):

PR #4457, opened 2023-03-22
Add BT only note to regional

### LINKED ISSUE #4448: Add plan availability (Free, Add-on, BT)on Regional Segment Page

https://segment.com/docs/guides/regional-segment/

Give as much detail as you can to help us understand the change you want to see.
Add the banner showing regional segment is available for BT customers etc (similar to the public API Docs)

Why should the docs be changed? What use cases does it support?
It will avoid confusion for customers, and may result in less tickets being created asking why they cannot create a EU workspace.

What is the expected outcome?
Customers understanding they need to upgrade in order to avail of this feature.

### Additional information
Screenshot 2023-03-20 at 11 51 20
Screenshot 2023-03-20 at 11 47 11

_No response_

markzegarelli:
### Related issues (optional)
Closes #4448

### DIFF
--- a/src/_data/products.yml
+++ b/src/_data/products.yml
@@ -138,7 +138,7 @@ items:
     business: true
     add-on: true
 - product_display_name: Regional Segment
-  slug: data-residency
+  slug: regional
   plan-note: "Regional Segment is available to customers on the Business Tier plan."
   plans:
     free: false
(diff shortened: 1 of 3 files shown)


forstisabella (approved):

forstisabella:
@markzegarelli is this gtg?

PR #4472, opened 2023-03-24
Update Ripe docs

simpajj:
### Proposed changes

Added new documentation and some more information

### Merge timing
ASAP once approved

### DIFF
--- a/src/connections/destinations/catalog/actions-ripe-cloud/index.md
+++ b/src/connections/destinations/catalog/actions-ripe-cloud/index.md
@@ -1,38 +1,99 @@
 ---
-title: Ripe Cloud (Actions) Destination
+title: Ripe Cloud Mode Destination
 hide-boilerplate: true
 hide-dossier: true
 private: true
 id: 63cade592992cf7052ce2e3e
 ---
 
-[Ripe](https://www.getripe.com/){:target="_blank"} is a product-led sales 
-platform that empowers you to unlock revenue pipeline with product data. By identifying and showing which prospects to focus efforts on, you can convert leads into meetings inside your product.
+[Ripe](https://www.getripe.com/){:target="_new"} is a sales conversion tool that enables B2B revenue teams to surface and meet their best leads at the best possible time - when they are using your product.
 
-This destination enables you to send product data to Ripe. Sales teams can identify who decision-makers and product champions are by understanding what properties they have and what events they have triggered. The Ripe destination is built as an alternative to directly adding Ripe’s SDK script to your app or site.
+The Ripe Segment integration is an [Actions-based Destination in cloud mode](https://segment.com/docs/connections/destinations/#connection-modes) that lets you send your product events data to Ripe.
 
-The Ripe Segment integration is an [Actions-based Destination in cloud mode](/docs/connections/destinations/#connection-modes) that lets you send your backend events directly to Ripe.
+Ripe maintains this destination. For any issues with the destination, [contact the Ripe team](mailto:support@getripe.com).
 
-{% include content/ajs-upgrade.md %}
+> success ""
+> Set up a free account with Ripe by visiting their [website](https://www.getripe.com/){:target="_new"}.
 
-## Benefits of Ripe
+## **Getting Started**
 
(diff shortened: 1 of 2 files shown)


markzegarelli:
Hi @simpajj, can you provide a bit of context on these updates? Docs with the content you've added are meant for our classic destinations, whereas the Ripe destinations use the Actions framework.

I realize there may have been a bit of time where the docs did not populate the metadata properly, but we resolved this earlier this week.

Thanks!

simpajj:
> Hi @simpajj, can you provide a bit of context on these updates? Docs with the content you've added are meant for our classic destinations, whereas the Ripe destinations use the Actions framework.
> 
> I realize there may have been a bit of time where the docs did not populate the metadata properly, but we resolved this earlier this week.
> 
> Thanks!

Hi @markzegarelli, thanks for getting back to me. This is mostly a copy overhaul of the documentation for both of our Actions destinations.

> Docs with the content you've added are meant for our classic destinations, whereas the Ripe destinations use the Actions framework.

My intention was to update the docs for the Actions destinations, but I've probably made some error here. Just so that I understand, is it the title that determines whether the documentation is for a Classic or Actions destination? Slightly confused as we don't have a Classic destination.

Thanks!

jdykert:
@markzegarelli wanted to create some more clarity on our objectives, hopefully this helps:

1. Rename destination naming to better reflect how we talk to our customers about them, the _(Actions)_ label seems superfluous since we don't have any classic integration out and thus have no need to differentiate between the old and the new. You [Segment] also do a great job of pointing out which destinations are action-enabled in your catalog UI. We also did not want to add the _web mode_ label to the web destination since that does not map to (or feels suitable) given how we will display it in the public catalog.

2. Call out the add-on nature of the cloud mode destination. While it's possible to use it as a standalone in theory, to enable the full Ripe feature set you need to enable our web mode destination for a JS source.

3. Help our mutual customers understand the specific calls and how they work with our default/OOB mapping for 
our two actions destination. We still include the Actions info in the docs (first paragraph as well as the last few lines for both destinations).

Would love your input on how to better document the _Actions_ capabilities for both entries!

markzegarelli:
Thanks @simpajj and @jdykert for the context! 

For the name change, please connect with Customer Success if you haven't yet. To be honest, I'm not sure if `(Actions)` is a requirement for destinations that use the Actions framework, but CS should be able to advise. We'll update the name on the article to match what we see in the back end.

For the content, the part I'm unsure about is documenting the JS calls for the supported methods. This is how we've documented classic destinations in the past. For actions though, mappings take place in the UI. Best practice in this case would be to add this content to the action descriptions, so that we can place it in context. For example, we would replace `Identify users in Ripe` [here](https://segment.com/docs/connections/destinations/catalog/actions-ripe-cloud/#identify) with:
```
Segment sends Identify calls to Ripe as `identify` events. Ripe displays these events as `Users` by default. `Identify` event data is augmented with traits. It's important to include email as a trait, as soon as it is known. Ripe will use email to populate User views by default. Ripe supports and follows the [Segment's Identify Spec](https://segment.com/docs/connections/spec/identify/#traits), and recommends using the standardized names for the reserved traits covered there.
```

jdykert:
@markzegarelli we're happy to do it that way instead. Can you share any additional boilerplates that would apply to the other sections?

markzegarelli:
Hi @jdykert, I'm not sure what you mean with regards to boilerplates for other sections. Is the end goal to get information like [this](https://github.com/segmentio/segment-docs/pull/4472/files#diff-1499c23bf926114f6d3df896cef8d740a6656d423ea95b8cec0401083b3537b4R46) into the doc to supplement what is there by default?

jdykert (approved):
Looks great!

markzegarelli:
@jdykert Thanks. We are working on a cleaner way to get that information into the docs, so it can bet contextually relevant with how the docs for Actions-based destinations are built now. Apologies for the delay, but we think this will make updates much smoother in the future. Please stay tuned.

cc: @joe-ayoub-segment - This is another use case for putting docs notes in the context of actions.

PR #4473, opened 2023-03-24
Update user-deletion-and-suppression.md

sneha-shashidhar:
### Proposed changes

Customers have written to Support with confusions related to the completion of suppression requests. Since the doc mentions 24 hours is the SLA to add a userId to the list of Suppressed Users, they assume that the users will be suppressed within that period. However, this is not the case and Segment takes up to 30 days.

### DIFF
--- a/src/privacy/user-deletion-and-suppression.md
+++ b/src/privacy/user-deletion-and-suppression.md
@@ -62,7 +62,7 @@ The Suppressed Users tab shows an up-to-date list of **actively** suppressed `us
 
 To create a suppression regulation and add a `userId` to this list, click **Suppress New User**, and enter the `userId` in the field that appears. Then click **Request Suppression**.
 
-Segment creates a `SUPPRESS` regulation, and adds the `userId` to your suppression list within 24 hours. But the suppression request can take up to 30 days to process. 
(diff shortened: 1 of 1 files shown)


src/privacy/user-deletion-and-suppression.md
```diff
@@ -62,7 +62,7 @@ The Suppressed Users tab shows an up-to-date list of **actively** suppressed `us
 
 To create a suppression regulation and add a `userId` to this list, click **Suppress New User**, and enter the `userId` in the field that appears. Then click **Request Suppression**.
 
-Segment creates a `SUPPRESS` regulation, and adds the `userId` to your suppression list within 24 hours.
+Segment creates a `SUPPRESS` regulation, and adds the `userId` to your suppression list within 24 hours. But the suppression request can take up to 30 days to process.
```
rchinn-segment, line 65:
```suggestion
Segment creates a `SUPPRESS` regulation, and adds the `userId` to your suppression list within 24 hours. Note that the suppression request can take up to 30 days to process. 
```

rchinn-segment (changes requested):

rchinn-segment (approved):

PR #4480, opened 2023-03-27
Add flush policy docs

niallzato:
### Proposed changes

Add flush policy docs, copied from here: https://github.com/segmentio/analytics-react-native#controlling-upload-with-flush-policies

### DIFF
--- a/src/connections/sources/catalog/libraries/mobile/react-native/index.md
+++ b/src/connections/sources/catalog/libraries/mobile/react-native/index.md
@@ -361,7 +361,7 @@ We have several standard FlushPolicies:
 - `StartupFlushPolicy` triggers on client startup only
 
 ### Adding or removing policies
-One of the main advantages of FlushPolicies is that you can add and remove policies on the fly. This is very powerful when you want to reduce or increase the amount of flushes. 
+One of the main advantages of Flush Policies is that you can add and remove policies on the fly. This is very powerful when you want to reduce or increase the amount of flushes. 
 For example you might want to disable flushes if you detect the user has no network:
 ```ts
 import NetInfo from "@react-native-community/netinfo";
--- a/src/connections/sources/catalog/libraries/mobile/react-native/index.md
+++ b/src/connections/sources/catalog/libraries/mobile/react-native/index.md
@@ -377,7 +377,7 @@ const client = createClient({
 // If we detect the user disconnects from the network remove all flush policies, 
 // That way the Segment client won't keep attempting to send events to Segment but will still 
 // store them for future upload.
-// If the network comes back up we add the policies back
+// If the network comes back up, the Segment client adds the policies back. 
 const unsubscribe = NetInfo.addEventListener((state) => {
   if (state.isConnected) {
     client.addFlushPolicy(...policiesIfNetworkIsUp);
--- a/src/connections/sources/catalog/libraries/mobile/react-native/index.md
+++ b/src/connections/sources/catalog/libraries/mobile/react-native/index.md
@@ -375,7 +375,7 @@ const client = createClient({
   flushPolicies: policies,
 });
 // If we detect the user disconnects from the network remove all flush policies, 
-// that way we won't keep attempting to send events to segment but we will still 
+// That way the Segment client won't keep attempting to send events to Segment but will still 
 // store them for future upload.
 // If the network comes back up we add the policies back
 const unsubscribe = NetInfo.addEventListener((state) => {
--- a/src/connections/sources/catalog/libraries/mobile/react-native/index.md
+++ b/src/connections/sources/catalog/libraries/mobile/react-native/index.md
@@ -340,7 +340,7 @@ If you don't do this, the old client instance would still exist and retain the t
 Ideally, you shouldn't have to use this method, and the Segment client should be initialized only once in the application lifecycle.
 
 ## Control upload with Flush Policies
-To granularly control when events are uploaded you can use `FlushPolicies`.
+To granularly control when Segment uploads events you can use `FlushPolicies`.
 A Flush Policy defines the strategy for deciding when to flush, this can be on an interval, on a certain time of day, after receiving a certain number of events or even after receiving a particular event. This gives you even more flexibility on when to send event to Segment.
 To make use of flush policies you can set them in the configuration of the client:
 ```ts
--- a/src/connections/sources/catalog/libraries/mobile/react-native/index.md
+++ b/src/connections/sources/catalog/libraries/mobile/react-native/index.md
@@ -390,7 +390,7 @@ const unsubscribe = NetInfo.addEventListener((state) => {
 You can create a custom Flush Policy special for your application needs by implementing the  `FlushPolicy` interface. You can also extend the `FlushPolicyBase` class that already creates and handles the `shouldFlush` value reset.
 A `FlushPolicy` only needs to implement two methods:
 - `start()`: Executed when the flush policy is enabled and added to the client. This is a good place to start background operations, make async calls, configure things before execution
-- `onEvent(event: SegmentEvent)`: Gets called on every event tracked by your client
+- `onEvent(event: SegmentEvent)`: Called on every event tracked by your client
 - `reset()`: Called after a flush is triggered (either by your policy, by another policy, or manually)
 Your policies also have a `shouldFlush` observable boolean value. When this is set to true the client attempts to upload events. Each policy should reset this value to `false` according to its own logic, although it's common to do it inside the `reset` method.
 ```ts
--- a/src/connections/sources/catalog/libraries/mobile/react-native/index.md
+++ b/src/connections/sources/catalog/libraries/mobile/react-native/index.md
@@ -353,7 +353,7 @@ const client = createClient({
   ],
 });
 ```
-You can set several policies at a time. Whenever any of them decides it is time for a flush it will trigger an upload of the events. The rest get reset so that their logic restarts after every flush. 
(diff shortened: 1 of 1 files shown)


src/connections/sources/catalog/libraries/mobile/react-native/index.md
```diff
@@ -339,6 +339,76 @@ If you don't do this, the old client instance would still exist and retain the t
...
+const unsubscribe = NetInfo.addEventListener((state) => {
+  if (state.isConnected) {
+    client.addFlushPolicy(...policiesIfNetworkIsUp);
+  } else {
+    client.removeFlushPolicy(...policiesIfNetworkIsUp)
+  }
+});
+```
+### Creating your own flush policies
+You can create a custom FlushPolicy special for your application needs by implementing the  `FlushPolicy` interface. You can also extend the `FlushPolicyBase` class that already creates and handles the `shouldFlush` value reset.
```
rchinn-segment, line 390:
```suggestion
You can create a custom FlushPolicy for your application needs by implementing the  `FlushPolicy` interface. You can also extend the `FlushPolicyBase` class that already creates and handles the `shouldFlush` value reset.
```

src/connections/sources/catalog/libraries/mobile/react-native/index.md
```diff
@@ -339,6 +339,76 @@ If you don't do this, the old client instance would still exist and retain the t
...
+A `FlushPolicy` only needs to implement 2 methods:
```
rchinn-segment, line 391:
```suggestion
A `FlushPolicy` only needs to implement two methods:
```

src/connections/sources/catalog/libraries/mobile/react-native/index.md
```diff
@@ -339,6 +339,76 @@ If you don't do this, the old client instance would still exist and retain the t
...
+- `start()`: Executed when the flush policy is enabled and added to the client. This is a good place to start background operations, make async calls, configure things before execution
+- `onEvent(event: SegmentEvent)`: Gets called on every event tracked by your client
```
rchinn-segment, line 393:
```suggestion
- `onEvent(event: SegmentEvent)`: Called on every event tracked by your client
```

src/connections/sources/catalog/libraries/mobile/react-native/index.md
```diff
@@ -339,6 +339,76 @@ If you don't do this, the old client instance would still exist and retain the t
...
+- `reset()`: Called after a flush is triggered (either by your policy, by another policy or manually)
```
rchinn-segment, line 394:
```suggestion
- `reset()`: Called after a flush is triggered (either by your policy, by another policy, or manually)
```

src/connections/sources/catalog/libraries/mobile/react-native/index.md
```diff
@@ -339,6 +339,76 @@ If you don't do this, the old client instance would still exist and retain the t
...
+They also have a `shouldFlush` observable boolean value. When this is set to true the client will attempt to upload events. Each policy should reset this value to `false` according to its own logic, although it is pretty common to do it inside the `reset` method.
```
rchinn-segment, line 395:
```suggestion
Your policies also have a `shouldFlush` observable boolean value. When this is set to true the client attempts to upload events. Each policy should reset this value to `false` according to its own logic, although it's common to do it inside the `reset` method.
```

src/connections/sources/catalog/libraries/mobile/react-native/index.md
```diff
@@ -339,6 +339,76 @@ If you don't do this, the old client instance would still exist and retain the t
...
+  flushPolicies: [
+    new CountFlushPolicy(5),
+    new TimerFlushPolicy(500),
+    new StartupFlushPolicy(),
+  ],
+});
+```
+You can set several policies at a time. Whenever any of them decides it is time for a flush it will trigger an upload of the events. The rest get reset so that their logic restarts after every flush. 
+That means only the first policy to reach `shouldFlush` gets to trigger a flush at a time. In the example above either the event count gets to 5 or the timer reaches 500ms, whatever comes first will trigger a flush.
+We have several standard FlushPolicies:
```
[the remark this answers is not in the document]
rchinn-segment: Not sure if it needs to be `FlushPolicies`, as it's written as `Flush Policies` above. Just want to try and make this as consistent as possible.

src/connections/sources/catalog/libraries/mobile/react-native/index.md
```diff
@@ -339,6 +339,76 @@ If you don't do this, the old client instance would still exist and retain the t
...
+const policiesIfNetworkIsUp = [
+  new CountFlushPolicy(5),
+  new TimerFlushPolicy(500),
+];
+// Create our client with our policies by default
+const client = createClient({
+  // ...
+  flushPolicies: policies,
+});
+// If we detect the user disconnects from the network remove all flush policies,
```
rchinn-segment, line 377: Do the instances of `we` in lines 377-380 all refer to `Segment`?

src/connections/sources/catalog/libraries/mobile/react-native/index.md
```diff
@@ -339,6 +339,76 @@ If you don't do this, the old client instance would still exist and retain the t
...
+```
+You can set several policies at a time. Whenever any of them decides it is time for a flush it will trigger an upload of the events. The rest get reset so that their logic restarts after every flush. 
+That means only the first policy to reach `shouldFlush` gets to trigger a flush at a time. In the example above either the event count gets to 5 or the timer reaches 500ms, whatever comes first will trigger a flush.
+We have several standard FlushPolicies:
+- `CountFlushPolicy` triggers whenever a certain number of events is reached
+- `TimerFlushPolicy` triggers on an interval of milliseconds
+- `StartupFlushPolicy` triggers on client startup only
+
+### Adding or removing policies
+One of the main advantages of FlushPolicies is that you can add and remove policies on the fly. This is very powerful when you want to reduce or increase the amount of flushes.
```
rchinn-segment, line 364:
```suggestion
One of the main advantages of Flush Policies is that you can add and remove policies on the fly. This is very powerful when you want to reduce or increase the amount of flushes. 
```

src/connections/sources/catalog/libraries/mobile/react-native/index.md
rchinn-segment, line 390:
```suggestion
You can create a custom Flush Policy special for your application needs by implementing the  `FlushPolicy` interface. You can also extend the `FlushPolicyBase` class that already creates and handles the `shouldFlush` value reset.
```

rchinn-segment (changes requested):
Hey @niallzato - thanks for submitting this. I've left some comments (and a few questions) to align this a bit more with our [style guide](https://github.com/segmentio/segment-docs/blob/develop/styleguide.md). After that, we'll add this to our docs!

src/connections/sources/catalog/libraries/mobile/react-native/index.md
niallzato: @rchinn-segment thanks for all the suggestions. Correct, we are referring to Segment for these instances. Specifically, the Segment library initialized in the app. We could say something like the [following](https://segment.com/docs/connections/sources/catalog/libraries/mobile/react-native/#:~:text=method%2C%20and%20the-,Segment%20client,-should%20be%20initialized), perhaps:

`// that way the Segment client won't keep attempting to send events to Segment, but they will still 
// be stored for future upload.
// If the network comes back up, the Segment client adds the policies back`

src/connections/sources/catalog/libraries/mobile/react-native/index.md
```diff
@@ -339,6 +339,76 @@ If you don't do this, the old client instance would still exist and retain the t
...
+  new CountFlushPolicy(5),
+  new TimerFlushPolicy(500),
+];
+// Create our client with our policies by default
+const client = createClient({
+  // ...
+  flushPolicies: policies,
+});
+// If we detect the user disconnects from the network remove all flush policies, 
+// that way we won't keep attempting to send events to segment but we will still
```
rchinn-segment, line 378:
```suggestion
// That way the Segment client won't keep attempting to send events to Segment but will still 
```

src/connections/sources/catalog/libraries/mobile/react-native/index.md
```diff
@@ -339,6 +339,76 @@ If you don't do this, the old client instance would still exist and retain the t
...
+];
+// Create our client with our policies by default
+const client = createClient({
+  // ...
+  flushPolicies: policies,
+});
+// If we detect the user disconnects from the network remove all flush policies, 
+// that way we won't keep attempting to send events to segment but we will still 
+// store them for future upload.
+// If the network comes back up we add the policies back
```
rchinn-segment, line 380:
```suggestion
// If the network comes back up, the Segment client adds the policies back. 
```

src/connections/sources/catalog/libraries/mobile/react-native/index.md
```diff
@@ -339,6 +339,76 @@ If you don't do this, the old client instance would still exist and retain the t
...
+const policiesIfNetworkIsUp = [
+  new CountFlushPolicy(5),
+  new TimerFlushPolicy(500),
+];
+// Create our client with our policies by default
+const client = createClient({
+  // ...
+  flushPolicies: policies,
+});
+// If we detect the user disconnects from the network remove all flush policies,
```
rchinn-segment, line 377:
```suggestion
// If Segment detects the user disconnect from the network, Segment removes all flush policies. 
```

PR #4493, opened 2023-03-29
Info about settings variable.

Panandhan22:
### Proposed changes

> info ""
> settings can't be accessed as a global variable in the code. It has to be passed to the function at runtime.

### DIFF
--- a/src/connections/functions/destination-functions.md
+++ b/src/connections/functions/destination-functions.md
@@ -159,7 +159,7 @@ async function onTrack(request, settings) {
 When you deploy your destination function in your workspace, you fill out the settings on the destination configuration page, similar to how you would configure a normal destination.
 
 > info ""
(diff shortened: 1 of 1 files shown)


src/connections/functions/destination-functions.md
```diff
@@ -158,6 +158,9 @@ async function onTrack(request, settings) {
 
 When you deploy your destination function in your workspace, you fill out the settings on the destination configuration page, similar to how you would configure a normal destination.
 
+> info ""
+> The settings object can't be accessed as a global variable in the code. It has to be passed to the function at runtime.
```
markzegarelli, line 162:
```suggestion
> You must pass the settings object to the function at runtime. Functions can't access the settings object when it's stored as a global variable.
```

markzegarelli (changes requested):

markzegarelli (approved):

PR #4510, opened 2023-03-31
Update two topics to clarify current behavior

prigiattiperrut:
### Proposed changes

Added more details on how to override context fields for page calls. 
Clarified the current behavior for when events exceed the size limit.

### Merge timing
- ASAP once approved

### DIFF
--- a/src/connections/sources/catalog/libraries/website/javascript/troubleshooting.md
+++ b/src/connections/sources/catalog/libraries/website/javascript/troubleshooting.md
@@ -76,7 +76,7 @@ The JavaScript console reveals all requests, outbound and inbound, to your brows
 
 ## Is there a size limit on requests?
 
-Yes, 32KB per event message. Events with a payload larger than 32KB are dropped, and Segment's servers respond with a `500` error.
+Yes, 32KB per event message. Events with a payload larger than 32KB are accepted by Analytics.js with the browser getting responses `200` from our servers, but the event will be silently dropped once it enters our pipeline. 
 
 ## If Analytics.js fails to load, are callbacks not fired?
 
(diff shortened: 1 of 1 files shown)


src/connections/sources/catalog/libraries/website/javascript/troubleshooting.md
```diff
@@ -76,7 +76,7 @@ The JavaScript console reveals all requests, outbound and inbound, to your brows
 
 ## Is there a size limit on requests?
 
-Yes, 32KB per event message. Events with a payload larger than 32KB are dropped, and Segment's servers respond with a `500` error.
+Yes, 32KB per event message. Events with a payload larger than 32KB are accepted by Analytics.js with the browser getting responses `200` from our servers, but the event will be silently dropped once it enters our pipeline.
```
forstisabella, line 79:
```suggestion
Yes, 32KB per event message. Events with a payload larger than 32KB are accepted by Analytics.js, with the browser returning a `200` response, but the event will be silently dropped once it enters Segment's pipeline. 
```

src/connections/sources/catalog/libraries/website/javascript/troubleshooting.md
```diff
@@ -130,6 +130,18 @@ analytics.track("Receipt Viewed", {}, {
 ```
 This works for any [context field](/docs/connections/spec/common/#context) that Segment automatically collects.
 
+When working with Page calls, you can overwrite context fields by following the same instructions above. However, because the `context.page` fields are also available in the `properties` parameter for page calls, you can also prevent the same fields in the `properties` parameter from being included in your Page call. The example below will allow you to overwrite `url` available in context field `page.url` and properties parameter:
```
forstisabella, line 133:
```suggestion
When working with Page calls, you can overwrite context fields by following the same instructions above. However, because the `context.page` fields are also available in the `properties` parameter for page calls, you must also prevent the same fields in the `properties` parameter from being included in your Page call. The example below will allow you to overwrite `url` available in context field `page.url` and properties parameter:
```

forstisabella (approved):

PR #4531, opened 2023-04-04
Add to FAQ on SQL traits returning arrays

bobbyatsegment:
### Proposed changes

SQL traits docs do not currently mention that returning arrays via SQL traits is not supported. Added info in FAQ section that pulling in array objects via SQL traits is not supported, along with an alternative option to achieve a similar result.

### DIFF
--- a/src/engage/audiences/sql-traits.md
+++ b/src/engage/audiences/sql-traits.md
@@ -205,7 +205,7 @@ Yes, Segment limits request sizes to a maximum of 16KB. Records larger than this
 
 ### Do SQL Traits support arrays?
 
-It is not possible to pull in an array object with SQL Traits, only strings and numeric traits are supported. Arrays can instead be cast as a comma-separated string - if you were to use this later while building an audience, you would then be able to check whether the array contains a certain value by using the “contains” operator, but the value would be sent to any connected destinations as a string.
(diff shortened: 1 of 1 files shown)


src/engage/audiences/sql-traits.md
```diff
@@ -203,6 +203,10 @@ If you're importing a large list of users and traits, you'll need to consider yo
 
 Yes, Segment limits request sizes to a maximum of 16KB. Records larger than this are discarded.
 
+### Do SQL Traits support arrays?
+
+It is not possible to pull in an array object with SQL Traits, only strings and numeric traits are supported. Arrays can instead be cast as a comma-separated string - if you were to use this later while building an audience, you would then be able to check whether the array contains a certain value by using the “contains” operator, but the value would be sent to any connected destinations as a string.
```
markzegarelli, line 208:
```suggestion
No, SQL Traits supports string and numeric data types. You can cast arrays as a comma-separated string. In this case, if you used this trait to build an audience, you could check if the array contains a certain value with the "contains" operator, but the value is sent to any connected destinations as a string.
```

markzegarelli (changes requested):
Looks good @bobbyatsegment!

Made a few edits for conciseness and clarity.

markzegarelli (approved):

PR #4533, opened 2023-04-05
Mixpanel requirement on sending anonymousID in UUID V4 format.

Panandhan22:
### Proposed changes

Mixpanel requirement on sending anonymousID in UUID V4 format.

> info ""
> Mixpanel does have a specific requirement for the anonymousId. Using Analytics.js library to generate anonymousIds, there should not be a problem once identifying a user with a userId. And if the anonymousIds are manually assigned from your end then it should be sent with UUID v4 format. All Segment assigned anonymousIds are in UUID V4 format.

### DIFF
--- a/src/connections/destinations/catalog/actions-mixpanel/index.md
+++ b/src/connections/destinations/catalog/actions-mixpanel/index.md
@@ -113,8 +113,8 @@ The group id that Mixpanel will use is `12345`.
 
 {% include components/actions-fields.html content1=track_purchase_details section1="trackPurchase" content2=group_identify_user_details section2="groupIdentifyUser" content3=identify_user_details section3="identifyUser" %}
 
-> info ""
-> [Mixpanel](https://developer.mixpanel.com/reference/create-identity#create-identity) does have a specific requirement for the anonymousId. Using Analytics.js library to generate anonymousIds, there should not be a problem once identifying a user with a userId. And if the anonymousIds are manually assigned then it should be sent with UUID v4 format. All Segment assigned anonymousIds are in UUID V4 format.
+> info "Anonymous ID format"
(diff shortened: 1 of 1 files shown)


src/connections/destinations/catalog/actions-mixpanel/index.md
```diff
@@ -112,6 +112,10 @@ The group id that Mixpanel will use is `12345`.
 {% endcapture %}
 
 {% include components/actions-fields.html content1=track_purchase_details section1="trackPurchase" content2=group_identify_user_details section2="groupIdentifyUser" content3=identify_user_details section3="identifyUser" %}
+
+> info ""
+> [Mixpanel](https://developer.mixpanel.com/reference/create-identity#create-identity) does have a specific requirement for the anonymousId. Using Analytics.js library to generate anonymousIds, there should not be a problem once identifying a user with a userId. And if the anonymousIds are manually assigned then it should be sent with UUID v4 format. All Segment assigned anonymousIds are in UUID V4 format.
```
markzegarelli, line 117:
```suggestion
> [Mixpanel](https://developer.mixpanel.com/reference/create-identity#create-identity){:target="_blank"} requires that values it receives for the anonymous identifier (`anonymousId` in Segment) must be in the UUID v4 format. Analytics.js sends `anonymousId` in this format by default. If you manually send anonymous identifiers to Mixpanel, ensure they are in the correct format.
```

src/connections/destinations/catalog/actions-mixpanel/index.md
```diff
@@ -112,6 +112,10 @@ The group id that Mixpanel will use is `12345`.
 {% endcapture %}
 
 {% include components/actions-fields.html content1=track_purchase_details section1="trackPurchase" content2=group_identify_user_details section2="groupIdentifyUser" content3=identify_user_details section3="identifyUser" %}
+
+> info ""
```
markzegarelli, line 116:
```suggestion
> info "Anonymous ID format"
```

markzegarelli (changes requested):

markzegarelli (approved):

PR #4544, opened 2023-04-06
Algolia insights update docs

beatrice94:
### Proposed changes

After making changes to our Segment Destination, we're updating the documentation to match.
We've removed and reworded some sections that were no longer relevant or confusing for the reader, and added details about the presets that are now included.

### Merge timing
- This should be published as soon as the destination is available to users


beatrice94:
@joe-ayoub-segment just mentioning you to keep you updated!

PR #4575, opened 2023-04-13
Include a FAQ & Troubleshooting section

Tinu-ops:
We get quite a few requests that are similar to this and believe that including this common error message will create more transparency for clients to understand the issue and how to resolve it. A customer also responded back to a csat mentioning how beneficial it would be for them in the docs; ” That was the perfect answer and should be included in the FAQ of the Destination (with the ID stripped out)”. This is the 
”From the error provided by FB, this appears to be an issue where the Pixel ID provided does not exist or there is a permissions' problem.

Unsupported post request. Object with ID 'XXXX' does not exist, cannot be loaded due to missing permissions, or does not support this operation. Please read the Graph API documentation at https://developers.facebook.com/docs/graph-api\


stayseesong:
@Tinu-ops is this meant to still be in draft mode?

PR #4611, opened 2023-04-20
Update Transfer To Workspace section w/ limitation

daldeano-segment:
### Proposed changes

Updated this [section](https://segment.com/docs/guides/usage-and-billing/account-management/#can-i-move-a-source-from-one-workspace-to-another) of our public documentation in order to make customers aware they aren't able to transfer sources to EU workspaces.  This is currently not supported, however, our UI doesn't prevent users from trying which has misled customers in the past.

### DIFF
--- a/src/guides/usage-and-billing/account-management.md
+++ b/src/guides/usage-and-billing/account-management.md
@@ -49,4 +49,4 @@ To move a source between workspaces, navigate to the source's **Settings** tab, 
 > Segment recommends that you disconnect Tracking Plans from Sources before you initiate a workspace transfer. Once the transfer is complete, add and reconnect your Tracking Plans in the new workspace.
 
 > warning "Sources can't be transferred to EU workspaces"
-> Though transferring sources to the EU workspace is not blocked in the UI, the transfer will not work as expected. This feature is currently unsupported for cross region migration. We recommend re-creating the source from scratch in the new workspace.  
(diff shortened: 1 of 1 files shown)


pmbasco (commented):
Hi @daldeano-segment - mostly looks good! But I see a warning that states `Avoid using first-person like 'We'. Use Segment, a product name, or the other company name.
`

So instead of '`We recommend re-creating the source from scratch in the new workspace.'`

Can you modify it so it says `'Segment recommends re-creating the source from scratch in the new workspace.' `

daldeano-segment:
Hey @pmbasco I modified the sentence to include "Segment" instead of "We".  Please let me know if anything else is needed!

src/guides/usage-and-billing/account-management.md
```diff
@@ -47,3 +47,6 @@ To move a source between workspaces, navigate to the source's **Settings** tab,
 
 > warning "Tracking Plans do not transfer"
 > Segment recommends that you disconnect Tracking Plans from Sources before you initiate a workspace transfer. Once the transfer is complete, add and reconnect your Tracking Plans in the new workspace.
+
+> warning "Sources can't be transferred to EU workspaces"
+> Though transferring sources to the EU workspace is not blocked in the UI, the transfer will not work as expected. This feature is currently unsupported for cross region migration. Segment recommends re-creating the source from scratch in the new workspace..
```
markzegarelli, line 52:
```suggestion
> Though transferring sources to the EU workspace is not blocked in the UI, the transfer will not work as expected. This feature is not supported for cross region migration. Segment recommends that you re-create the source in the new workspace. 
```

markzegarelli (approved):

PR #4620, opened 2023-04-21
add redirect, and remove hidden & private front matter

tcgilbert:
### Proposed changes

So the actual slug for this integration is `june-actions` as opposed to `actions-june`. I added a redirect, but if better to change folder name let me know!

### Merge timing
asap


rchinn-segment (approved):
Thanks @tcgilbert - I went ahead and changed the folder name in our repo. Will include this in our next docs deploy (4/25)!

PR #4651, opened 2023-04-27
GAclientId may cause user merge

Panandhan22:
### Proposed changes

The Google clientID(ga_clientid) is a unique value created for each browser-device pair and will exist for 2 years if the cookie is not cleared. The analytics.reset() call will be triggered from Segment end when the user logs off. This call will clear the cookies and local Storage created by Segment. It doesn’t clear data from other integrated tools. So on the next login, the user will be assigned with unique anonymous_id but with the same ga_clientid(as the cookie is not cleared). Hence, the profiles with different anonymous_id but with same ga_clientid will get merged.

### DIFF
--- a/src/unify/identity-resolution/externalids.md
+++ b/src/unify/identity-resolution/externalids.md
@@ -35,7 +35,7 @@ Segment automatically promotes the following traits and IDs in track and identif
 | ios.push_token     | context.device.token when context.device.type = 'ios'                                                         |
 
 > note ""
-> The Google clientID(ga_clientid) is a unique value created for each browser-device pair and will exist for 2 years if the cookie is not cleared. The analytics.reset() call will be triggered from Segment end when the user logs off. This call will clear the cookies and local Storage created by Segment. It doesn’t clear data from other integrated tools. So on the next login, the user will be assigned with unique anonymous_id but with the same ga_clientid(as the cookie is not cleared). Hence, the profiles with different anonymous_id but with same ga_clientid will get merged.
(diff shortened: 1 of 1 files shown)


src/unify/identity-resolution/externalids.md
```diff
@@ -34,6 +34,9 @@ Segment automatically promotes the following traits and IDs in track and identif
 | ios.idfa           | context.device.advertisingId when context.device.type = 'ios'     |
 | ios.push_token     | context.device.token when context.device.type = 'ios'                                                         |
 
+> note ""
+> The Google clientID(ga_clientid) is a unique value created for each browser-device pair and will exist for 2 years if the cookie is not cleared. The analytics.reset() call should be triggered from Segment end when the user logs off. This call will clear the cookies and local Storage created by Segment. It doesn’t clear data from other integrated tools. So on the next login, the user will be assigned with a new unique anonymous_id, but the same ga_clientid will remain if this cookie is not cleared. Hence, the profiles with different anonymous_id but with same ga_clientid will get merged.
```
rchinn-segment, line 38:
```suggestion
> The Google clientID(ga_clientid) is a unique value created for each browser-device pair and will exist for two years if you don't clear this cookie. When a user logs off, Segment triggers the analytics.reset() call which clears cookies and local storage Segment has created. However, analytics.reset() doesn’t clear data from other integrated tools. Therefore, when the user logs in again, Segment assigns a new unique anonymous_id, but the user will have the same ga_clientid. As a result, Segment merges these profiles, which have different anonymous_id but the same ga_clientid.
```

rchinn-segment (changes requested):
Thanks for submitting this, @Panandhan22 - left a few suggested changes to align this with the docs style guide.

markzegarelli (approved):

PR #4652, opened 2023-04-27
Add info about $value mapping for Klaviyo's Track events

joeynmq:
### Proposed changes
Checking our source code for Klaviyo device-mode destination: https://github.com/segmentio/analytics.js-integrations/blob/master/integrations/klaviyo/lib/index.js#L109-L117  

We map the `revenue` property from Segment's Track event to Klaviyo's `$value` field. 

So, when customers include a revenue property in Track events, they will see unexpected fields in Klaviyo:
image

### Merge timing
ASAP once approved

### DIFF
--- a/src/connections/destinations/catalog/klaviyo/index.md
+++ b/src/connections/destinations/catalog/klaviyo/index.md
@@ -142,6 +142,8 @@ analytics.track({
 
 When you call `track` on `analytics.js`, Segment calls Klaviyo's `track` with the same parameters.
 
+If you include `properties.revenue` in Track event, Segment maps it to Klaviyo's event `$value`.
+
 > info ""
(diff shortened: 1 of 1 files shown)


src/connections/destinations/catalog/klaviyo/index.md
```diff
@@ -142,6 +142,8 @@ analytics.track({
 
 When you call `track` on `analytics.js`, Segment calls Klaviyo's `track` with the same parameters.
 
+If you include `properties.revenue` in Track event, Segment maps it to Klaviyo's event `$value`.
```
rchinn-segment, line 145:
```suggestion
If you include `properties.revenue` in a track event, Segment maps it to Klaviyo's `$value` event.
```

rchinn-segment (approved):

PR #4690, opened 2023-05-08
feat: redesign react native documentation to match swift/kotlin [netlify-build]

alanjcharles:
### Proposed changes
re-designed the react native documentation to match the swift/kotlin redesign 

### Merge timing
 (hopefully) May 15, 2023


src/connections/sources/catalog/libraries/mobile/react-native/index.md
```diff
@@ -9,19 +9,12 @@ React Native 2.0 is a major version upgrade to the [existing React Native librar
...
 
 > info "Using Analytics for React Native Classic?"
-> If you're still using the classic version of Analytics for React Native, you can refer to the documentation [here](/docs/connections/sources/catalog/libraries/mobile/react-native/classic).
-> <br><br>On May 15, 2023, Segment will end support for Analytics React Native Classic, which includes versions 1.5.1 and older. [Upgrade to Analytics React Native 2.0](/docs/connections/sources/catalog/libraries/mobile/react-native/migration/). See the [Analytics React Native 2.0 docs](/docs/connections/sources/catalog/libraries/mobile/react-native/) to learn more. 
-
-Upgrade to Analytics React Native 2.0. See the Analytics React Native 2.0 docs to learn more.  
-
-If you're migrating to Analytics React Native 2.0 from an older Analytics React Native version, skip to the [migration guide](/docs/connections/sources/catalog/libraries/mobile/react-native/migration/).
+> Analytics React Native versions 1.5.1 and older are no longer supported by Segment. You can reference the [migration guide](/docs/connections/sources/catalog/libraries/mobile/react-native/migration/) to upgrade to the latest version.
```
stayseesong, line 12:
```suggestion
> As of May 15, 2023, Segment ended support for [Analytics React Native Classic](/docs/connections/sources/catalog/libraries/mobile/react-native/classic), which includes versions 1.5.1 and older. [Upgrade to Analytics React Native 2.0](/docs/connections/sources/catalog/libraries/mobile/react-native/migration/) to use the latest version.
```

stayseesong: @alanjcharles we need to include a link to the Classic docs as the classic version will still exist at the end of support date until all users upgrade to the new version. We also don't use the term "migration guide" as "upgrade" is a better term

stayseesong (changes requested):

src/connections/sources/catalog/libraries/mobile/react-native/index.md
alanjcharles: Hi @stayseesong we've decided as a team to remove all reference to the classic library in the documentation so that we don't need to call the new one 2.0 and rather just refer to it as `Analytics React Native`. With that said, we can still provide a link to the old one somewhere in the docs so people can access it if you think that's the way to go

markzegarelli (approved):

PR #4708, opened 2023-05-10
Updated typo

MStephen024:
### Proposed changes

Fixed a typo in the "Server-side Track" section of the docs and updated the description to be consistent with the "Client-side Track" description.

Currently: "When you call make a Track call from..."
Updated: "When you call `track` from..."

### DIFF
--- a/src/connections/destinations/catalog/klaviyo/index.md
+++ b/src/connections/destinations/catalog/klaviyo/index.md
@@ -149,7 +149,7 @@ If you include `properties.revenue` in a track event, Segment maps it to Klaviyo
 
 ### Server-side Track
 
-When you call `track` from one of Segment's mobile or server-side libraries, Segment keys the user with the `userId` and also provides the Klaviyo `$email` `customer_property` if your `userId` is an email, or you provide `email` as one of your event `properties`.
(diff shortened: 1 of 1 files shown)


src/connections/destinations/catalog/klaviyo/index.md
```diff
@@ -149,7 +149,7 @@ If you include `properties.revenue` in a track event, Segment maps it to Klaviyo
 
 ### Server-side Track
 
-When you call make a Track call from one of Segment's mobile or server-side libraries, Segment keys the user with the `userId` and also provides the Klaviyo `$email` `customer_property` if your `userId` is an email, or you provide `email` as one of your event `properties`.
+When you call `track` from one of Segment's mobile or server-side libraries, Segment keys the user with the `userId` and also provides the Klaviyo `$email` `customer_property` if your `userId` is an email, or you provide `email` as one of your event `properties`.
```
forstisabella, line 152: It was a good call to standardize the way we refer to calls, but we capitalize the calls when we talk about them ([style guide for reference](https://github.com/segmentio/segment-docs/blob/develop/styleguide.md#styling-the-segment-methods)). I'll make sure the client-side Track call gets updated to match.

```suggestion
When you make a Track call from one of Segment's mobile or server-side libraries, Segment keys the user with the `userId` and also provides the Klaviyo `$email` `customer_property` if your `userId` is an email, or you provide `email` as one of your event `properties`.
```

forstisabella (approved):

PR #4716, opened 2023-05-11
Update index.md

dineshraj-arunmozhi:
Updating the document to have more clear wording

### DIFF
--- a/src/connections/reverse-etl/index.md
+++ b/src/connections/reverse-etl/index.md
@@ -89,7 +89,7 @@ To add your first destination:
 4. Select the Reverse ETL source you want to connect the destination to.
 5. Enter the **Destination name** and click **Create Destination**.
 6. Enter the required information on the **Settings** tab of the destination.
-7. Enable the destination.
+7. Navigate to the destination settings tab and enable the destination. Note that, if the destination is disabled then Segment will not be able to start sync.
 
 ### Step 4: Create mappings
 After you’ve added a destination, you can create mappings from your warehouse to the destination. Mappings enable you to map the data you extract from your warehouse to the fields in your destination.


src/connections/reverse-etl/index.md
```diff
@@ -89,7 +89,7 @@ To add your first destination:
 4. Select the Reverse ETL source you want to connect the destination to.
 5. Enter the **Destination name** and click **Create Destination**.
 6. Enter the required information on the **Settings** tab of the destination.
-7. Enable the destination.
+7. Navigate to the destination settings tab and enable the destination. Note that, if the destination is disabled then Segment will not be able to start sync.
```
stayseesong, line 92:
```suggestion
7. Navigate to the destination settings tab and enable the destination. If the destination is disabled, then Segment won't be able to start sync.
```

stayseesong (changes requested):

src/connections/reverse-etl/index.md
```diff
@@ -89,7 +89,7 @@ To add your first destination:
 4. Select the Reverse ETL source you want to connect the destination to.
 5. Enter the **Destination name** and click **Create Destination**.
 6. Enter the required information on the **Settings** tab of the destination.
-7. Enable the destination.
+7. Navigate to the destination settings tab and enable the destination. If the destination is disabled, then Segment won't be able to start sync.
```
stayseesong, line 92: @dineshraj-arunmozhi curious why you want to add in that you have to navigate to the settings tab? From step 6, you should still be on the settings tab

dineshraj-arunmozhi: Hey @stayseesong, We have customers writing to us stating their sync has not started. While investigating, we noticed every time it has to do with the enabling the destination. Additionally, a couple of customers had felt that is not clear enough for them.

stayseesong (approved):

PR #4723, opened 2023-05-12
User Token is not taken for validation as Email is not in valid format.

Panandhan22:
### Proposed changes

Email and User Token both fields are mapped in "Send custom behavioural event" mapping and sent over the event paylaod too. But the event is failing for delivery with "Payload validation error" as the email is not in valid format, why it is not taking the User token for validation? Event payload should have the email with valid format or empty string, so that the user token will take precedence and can be validated. An invalid email will not pass mapping validation. Hence the event delivery is failing.


src/connections/destinations/catalog/actions-hubspot-cloud/index.md
```diff
@@ -65,3 +65,5 @@ HubSpot has several limits for custom behavioral events, including a limit on th
 ### Does the HubSpot Cloud Mode (Actions) destination support EU data residency?
 Yes. HubSpot will automatically redirect API requests directly to an EU data center if your HubSpot instance is on an EU data center. See more in HubSpot's [Routing API Traffic](https://product.hubspot.com/blog/routing-api-traffic){:target="_blank"} article.
 
+### How to attribute custom behavioural event using User token instead of Email?
```
rchinn-segment, line 68:
```suggestion
### How do I attribute a custom behavioral event with a user token instead of Email?
```

src/connections/destinations/catalog/actions-hubspot-cloud/index.md
```diff
@@ -65,3 +65,5 @@ HubSpot has several limits for custom behavioral events, including a limit on th
...
+Event payload should have the email with valid format or empty string or null, so that the user token will take precedence and can be validated in the "Send custom behavioural event" mapping. An invalid email will not pass mapping validation. And the event can't be delivered to the destination.
```
rchinn-segment, line 69:
```suggestion
Event payloads should contain an email with either a valid format, empty string, or `null`, so that the user token takes precedence and is validated in a `Send custom behavioral event` mapping. Segment can't deliver the event to your destination if the email is invalid.
```

rchinn-segment (changes requested):

rchinn-segment (approved):

PR #4742, opened 2023-05-16
Update firebase for React Native updates

niallzato:
### Proposed changes
Removing content not applicable to React Native, and added info around screens calls


alanjcharles (approved):
thanks man!

PR #4751, opened 2023-05-17
clarification on timestamp values - Intercom classic

spencerattick:
### Proposed changes

Segment does some work behind the scenes in a way that changes some event names to Intercom. This documentation makes that more clear.

### Merge timing
ASAP is fine

### DIFF
--- a/src/connections/destinations/catalog/intercom/index.md
+++ b/src/connections/destinations/catalog/intercom/index.md
@@ -96,7 +96,7 @@ When you call Identify, Segment creates or updates the user in Intercom using th
 
 Keep reading for more information about the Identify call depending on the source type you send it from.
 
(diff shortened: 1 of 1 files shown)


src/connections/destinations/catalog/intercom/index.md
```diff
@@ -96,6 +96,9 @@ When you call Identify, Segment creates or updates the user in Intercom using th
 
 Keep reading for more information about the Identify call depending on the source type you send it from.
 
+When you send a field into Intercom that has a timestamp as the value, Segment will convert the timestamp to UNIX and will append the suffix `_at` to the field name in order to tell Intercom that the field holds a timestamp value rather than a number. This is per Intercom's recommendation here: https://www.intercom.com/help/en/articles/3605703-how-dates-work-in-intercom.
```
markzegarelli, line 99:
```suggestion
When you send a field that has a timestamp as the value to Intercom, Segment converts the timestamp to the UNIX format and appends the suffix `_at` to the field name. This instructs Intercom that the field holds a timestamp value rather than a number. For more information, see [How Dates work in Intercom](https://www.intercom.com/help/en/articles/3605703-how-dates-work-in-intercom){:target="_blank"}. 
```

markzegarelli (changes requested):

markzegarelli (approved):

PR #4776, opened 2023-05-23
Update what-is-replay.md

sneha-shashidhar:
### Proposed changes

Customers often write in to check if it's possible to exclude certain records from being replayed based on some conditions. This is currently not supported and would be good to make a note of it in our public documentation.

### DIFF
--- a/src/guides/what-is-replay.md
+++ b/src/guides/what-is-replay.md
@@ -36,7 +36,7 @@ Replays do not affect your [MTU count](/docs/guides/usage-and-billing/mtus-and-t
 
 Once a replay starts, you will not see replayed events in the Event Delivery tab.
 
-Replays can be initiated for some or all events but it is not possible to apply conditional filters that excludes certain rows of data from being replayed. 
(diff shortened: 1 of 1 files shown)


src/guides/what-is-replay.md
```diff
@@ -36,6 +36,8 @@ Replays do not affect your [MTU count](/docs/guides/usage-and-billing/mtus-and-t
 
 Once a replay starts, you will not see replayed events in the Event Delivery tab.
 
+Replays can be initiated for some or all events but it is not possible to apply conditional filters that exclude certain rows of data from being replayed. You can set up [destination filters](/docs/connections/destinations/destination-filters/) to conditionally filter replayed events.
```
stayseesong, line 39:
```suggestion
You can initiate replays for some or all events, but you can't apply conditional filters that exclude certain rows of data from being replayed. You can set up [destination filters](/docs/connections/destinations/destination-filters/) to conditionally filter replayed events.
```

stayseesong (changes requested):

stayseesong (approved):

PR #4797, opened 2023-05-26
Update roles.md

sneha-shashidhar:
### Proposed changes

The Roles doc does not highlight the roles and permissions needed to access the Data Retention settings in End-user privacy. After testing it out, it seems that the End-user privacy access enables our customers to view the deletion and suppression lists but not data retention periods. A customer had this confusion and it would be beneficial to have this noted in the public docs.

### DIFF
--- a/src/segment-app/iam/roles.md
+++ b/src/segment-app/iam/roles.md
@@ -97,3 +97,7 @@ To **create or edit** transformations you must have either `Source Admin` for al
 ## Roles for Privacy Portal
 
 To **view, create or edit** PII related configurations in Privacy Portal, you need to have the `Workspace Owner` role.
+
+## Roles for Data Retention settings
+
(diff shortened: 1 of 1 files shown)


src/segment-app/iam/roles.md
```diff
@@ -96,4 +96,4 @@ To **create or edit** transformations you must have either `Source Admin` for al
 
 ## Roles for Privacy Portal
 
-To **view, create or edit** PII related configurations in Privacy Portal, you need to have the `Workspace Owner` role.
+The Privacy Portal is only accessible by `Workspace owners`. To **view, create or edit** any section of Privacy Portal, you will need to have the `Workspace Owner` role.
```
stayseesong, line 99:
```suggestion
The Privacy Portal is only accessible by `Workspace owners`. To **view, create or edit** any section of the Privacy Portal, you need to have the `Workspace Owner` role.
```

stayseesong (approved):

PR #4801, opened 2023-05-26
Update index.md repeater to include details on replay

lizkane222:
Add Replay details when using a Repeater

### Proposed changes

Added the section Replays with a Repeater
Replays can affect your MTU count when using a Repeater destination. Notify your team before initiating a Replay if you’re using a Repeater destination. There are edge cases where the repeater would count towards MTUs, specifically when replaying historical data from the source connected to feed data into that Repeater destination. This is because userIds/anonymousIds would not have been seen by the source the Repeater sends data to. The API plans count by events sent through the pipeline; the repeater resends an event through the entire pipeline [Source → Repeater → Source → Destinations] so one event is sent through twice, and would increase the throughput count.

This is necessary to add at this point in the docs so that customers are aware of the nuances associated with repeaters and replays.

### Merge timing
- once approved

### Related issues (optional)

Zendesk tickets : 
- https://segment.zendesk.com/agent/tickets/508182
- https://segment.zendesk.com/agent/tickets/361016


src/connections/destinations/catalog/repeater/index.md
```diff
@@ -26,6 +26,10 @@ You can do this for as many sources as you need.
 
 Repeater sends all events it receives to the sources you specified, identified by the write key(s) you added.
 
+## Replays with a Repeater
+
+Replays can affect your MTU count when using a Repeater destination. Notify your team before initiating a Replay if you’re using a Repeater destination. There are edge cases where the repeater would count towards MTUs, specifically when replaying historical data from the source connected to feed data into that Repeater destination. This is because userIds/anonymousIds would not have been seen by the source the Repeater sends data to. The API plans count by events sent through the pipeline; the repeater resends an event through the entire pipeline [Source → Repeater → Source → Destinations] so one event is sent through twice, and would increase the throughput count.
```
forstisabella, line 31:
```suggestion
Running a Replay on a Repeater destination might count toward your MTUs, especially if you are replaying historical data from the source that flows data into your Repeater destination. 

Because the API plans count by events sent through the pipeline and the Repeater destination resends an event through the entire pipeline, one event might flow through your source twice which increases the throughput count.

Segment recommends that you notify your team before initiating a Replay if you’re using a Repeater destination. 
```

forstisabella (approved):

PR #4808, opened 2023-05-27
Update actions.md Copying Classic Destination Filters to Actions Dest…

lizkane222:
…ination

Add section on Copying Classic Destination Filters to Actions Destination

### Proposed changes

Adding this section within our docs allows a more comprehensive understanding for how to migrate existing destination filters from their Classic destination to Actions destination.

### Merge timing
- ASAP once approved?

### Related issues (optional)

Zendesk Ticket : 
- https://segment.zendesk.com/agent/tickets/509512

### DIFF
--- a/src/connections/destinations/actions.md
+++ b/src/connections/destinations/actions.md
@@ -82,6 +82,16 @@ Moving from a classic destination to an actions-based destination is a manual pr
 5. Verify that data is flowing from the development or test source to the partner tool.
 6. Repeat the steps above with your production source.
 
+Copying over your Destination Filters from the Classic Destination to the Actions Destination
+
+This functionality is not available within the UI, but is available via the Public API for all customers on Segment's Business Tier Plan.
+1. Send a request to the Public API endpoint : [List Filters from Destination]([url](https://docs.segmentapis.com/tag/Destination-Filters#operation/listFiltersFromDestination)). The `destinationId` can be found in the URL while viewing the destination in your Segment workspace.
+  - List Filters from Destination : `GET` - `https://api.segmentapis.com/destination/{destinationId}/filters?pagination%5Bcount%5D=100`
+2. Grab the response and parse through the `data.filters` object. Each object returned inside the `data.filters` object is an individual filter associated with that Intercom Prod destination.
+4. Next, send individual `POST` requests to the Public API endpoint : [Create Filter for Destination]([url](https://docs.segmentapis.com/tag/Destination-Filters/#operation/createFilterForDestination)), for each of the filters from step 2. Specify the Actions `destinationId`, found in the URL when viewing that destination, and then the body of the request is the individual filters from step 2.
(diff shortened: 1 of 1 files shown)


src/connections/destinations/actions.md
```diff
@@ -82,6 +82,16 @@ Moving from a classic destination to an actions-based destination is a manual pr
...
+  - List Filters from Destination : `GET` - `https://api.segmentapis.com/destination/{destinationId}/filters?pagination%5Bcount%5D=100`
+2. Grab the response and parse through the `data.filters` object. Each object returned inside the `data.filters` object is an individual filter associated with that Intercom Prod destination.
+4. Next, send individual `POST` requests to the Public API endpoint : [Create Filter for Destination]([url](https://docs.segmentapis.com/tag/Destination-Filters/#operation/createFilterForDestination)), for each of the filters from step 2. Specify the Actions `destinationId`, found in the URL when viewing that destination, and then the body of the request is the individual filters from step 2.
+  - POST - https://api.segmentapis.com/destination/{destinationId}/filters
+6. If the bodies of those requests didn't already include the field `"enabled": true`, make sure to enable each of those filters after creating them.
```
stayseesong, line 93:
```suggestion
### Migrate your destination filters from the classic destination to the actions destination

> warning ""
> You can only migrate your destination filters using the Public API if you're on the business tier plan. This functionality isn't available in the Segment app. 

To migrate your destination filters to your actions destination from the classic destination: 
1. Send a request to the Public API endpoint. 
     - Use [List Filters from Destination](https://docs.segmentapis.com/tag/Destination-Filters#operation/listFiltersFromDestination). The `destinationId` can be found in the URL while viewing the destination in your Segment workspace.
2. Grab the response and parse through the `data.filters` object. Each object returned inside the `data.filters` object is an individual filter associated with that Intercom Prod destination.
4. Send individual `POST` requests to the Public API endpoint.
     - Use [Create Filter for Destination](https://docs.segmentapis.com/tag/Destination-Filters/#operation/createFilterForDestination), for each of the filters from step 2. 
     - Specify the Actions `destinationId`, found in the URL when viewing that destination. The body of the request is the individual filters from step 2.
6. If the bodies of those requests don't already include the field `"enabled": true`, make sure to enable each of those filters after you create them.
```

stayseesong: @lizkane222 I'm a little confused why "Intercom Prod destination" is mentioned in step 2. What if users don't use the Intercom prod destination?

stayseesong (changes requested):

src/connections/destinations/actions.md
lizkane222: @stayseesong so sorry it's taken so long to get back to you on this. 

You're absolutely correct, I must have copy/pasted that over and this should actually just say :
"Each object returned inside the `data.filters` object is an individual filter associated with that specified destination."

stayseesong (approved):

PR #4812, opened 2023-05-29
Add FAQ's on notifications to backgrounded Android apps

ashton-huxtable:
### Proposed changes

Customers can use Firebase Cloud messaging to implement Push Notifications. With the way FCM is designed, customers will receive the Push events in FCM even if the app is back grounded, but they will not reach Segment. I added a section that addresses why events do not reach Segment when an Android app is backgrounded, as well as one addressing why Firebase receives them even though Segment does not. 

### Merge timing
- ASAP once approved

### Related issues (optional)

n/a


niallzato (commented):
Looks great, one slight change needed I think!

pwseg (approved):

PR #4814, opened 2023-05-30
Add FAQ for data not reaching destination

bobbyatsegment:
### Proposed changes

Added a FAQ topic that covers how events may not be sending to the destination if mappings aren't enabled/configured to handle any events - in many instances, customers enabling destinations without mappings configured have written in to support expressing confusion on this, and it's mentioned in several destination docs (GA4 Cloud for example). Adding it in these docs as well for further visibility.

### DIFF
--- a/src/connections/destinations/actions.md
+++ b/src/connections/destinations/actions.md
@@ -187,6 +187,10 @@ You can combine criteria in a single group using **ALL** or **ANY**.  Use an ANY
 
 When you send an event with an actions destination Event Tester that doesn't match the trigger of any configured and enabled mappings, you'll see an error message that states, *You may not have any subscriptions that match this event.* To resolve the error, create a mapping with a trigger to handle the event being tested, or update the test event's payload to match the trigger of any existing mappings. 
 
+### Data not sending downstream
+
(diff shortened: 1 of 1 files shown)


src/connections/destinations/actions.md
```diff
@@ -187,6 +187,10 @@ You can combine criteria in a single group using **ALL** or **ANY**.  Use an ANY
 
 When you send an event with an actions destination Event Tester that doesn't match the trigger of any configured and enabled mappings, you'll see an error message that states, *You may not have any subscriptions that match this event.* To resolve the error, create a mapping with a trigger to handle the event being tested, or update the test event's payload to match the trigger of any existing mappings. 
 
+### Data not sending downstream
+
+Ensure that at least one mapping has been configured and enabled in the destination mappings for an event that you would like to reach downstream. If no mappings are enabled to trigger on an event that has been received from the connected source, the destination will not send any events.
```
forstisabella, line 192:
```suggestion
If no mappings are enabled to trigger on an event that has been received from the connected source, the destination will not send any events. Ensure that at least one mapping has been configured and enabled in the destination mappings for an event that you would like to reach downstream. 
```

forstisabella (approved):

PR #4839, opened 2023-06-02
Add FAQ for Other properties

bobbyatsegment:
### Proposed changes

Per this slack thread: https://twilio.slack.com/archives/CC97A542H/p1669907095535879?thread_ts=1669882189.824319&cid=CC97A542H

HubSpot (Actions) does not allow mapping an entire object to Other properties field because it helps to avoid passing a huge object of key/value pairs that will get rejected.

Adding this information to FAQs per customer request

### DIFF
--- a/src/connections/destinations/catalog/actions-hubspot-cloud/index.md
+++ b/src/connections/destinations/catalog/actions-hubspot-cloud/index.md
@@ -62,6 +62,10 @@ HubSpot has several limits for custom behavioral events, including a limit on th
 > note ""
 > A HubSpot Enterprise Marketing Hub account is required to send Custom Behavioral Events.
 
+### Why can't I set an entire object for the Other properties field?
+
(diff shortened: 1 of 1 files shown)


src/connections/destinations/catalog/actions-hubspot-cloud/index.md
```diff
@@ -62,6 +62,10 @@ HubSpot has several limits for custom behavioral events, including a limit on th
 > note ""
 > A HubSpot Enterprise Marketing Hub account is required to send Custom Behavioral Events.
 
+### Why can't I set an entire object for the Other properties field?
+
+This destination does not allow selecting an entire object for the Other properties field. HubSpot rejects API calls if a property name does not match with HubSpot's internal name. To prevent the possibility of rejection when working with a large object of key/value pairs, explicit mapping of each key/value pair is required. This helps to ensure that every key matches the pre-created property names in HubSpot.
```
rchinn-segment, line 67:
```suggestion
This destination doesn't allow selecting an entire object for the Other properties field. HubSpot rejects API calls if a property name doesn't match with HubSpot's internal name. When working with a large object of key/value pairs, map each key/value pair to prevent rejection. This ensures that every key matches the pre-created property names in HubSpot.
```

rchinn-segment (approved):

PR #4856, opened 2023-06-08
Multiple destination instance update

vanand17:
### Proposed changes

Add an information box about destinations that do not support multiple instances

### Merge timing
NA

### Related issues (optional)
NA

### DIFF
--- a/src/connections/destinations/add-destination.md
+++ b/src/connections/destinations/add-destination.md
@@ -114,6 +114,8 @@ Some destinations do not support having multiple instances connected to the same
 
 You can create unique destination filters for each destination instance connected to the same source.
 
+>info ""
+>Some destinations do not support having multiple instances connected to the same source. In that case, the option to add a second instance of that destination does not appear.
 
 ### Connect multiple sources to one instance of a destination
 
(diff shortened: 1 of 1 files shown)


src/connections/destinations/add-destination.md
```diff
@@ -114,6 +114,8 @@ Some destinations do not support having multiple instances connected to the same
 
 You can create unique destination filters for each destination instance connected to the same source.
 
+>info ""
+>Some destinations do not support having multiple instances connected to the same source. In that case, the option to add a second instance of that destination does not appear.
```
markzegarelli, line 118:
```suggestion
> Some destinations don't support multiple instances connected to the same source. If this is the case, you won't see the option to add a second instance of that destination.
```

src/connections/destinations/add-destination.md
```diff
@@ -114,6 +114,8 @@ Some destinations do not support having multiple instances connected to the same
 
 You can create unique destination filters for each destination instance connected to the same source.
 
+>info ""
```
markzegarelli, line 117:
```suggestion
> info ""
```

markzegarelli (approved):

PR #4861, opened 2023-06-08
Added a portion to FAQ to include modifying context object

bcaudillo:
### Proposed changes
I've had a few tickets where clients ask about modifying the context object, I thought it'd be good to have a plugin explicitly showing them. 

### Merge timing
Once approved

### DIFF
--- a/src/connections/sources/catalog/libraries/mobile/react-native/index.md
+++ b/src/connections/sources/catalog/libraries/mobile/react-native/index.md
@@ -693,7 +693,7 @@ import {
   }
 ```
 ### How do I add to the context Object?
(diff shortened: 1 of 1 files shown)


src/connections/sources/catalog/libraries/mobile/react-native/index.md
```diff
@@ -692,5 +692,34 @@ import {
     }
   }
 ```
+### How do I add to the context Object?
+Similar to the integrations object above you'll need to use a plugin to access and modify the context object. For example, if you'd like to add `context.groupId` to every Track call this can be added like so:
```
forstisabella, line 696:
```suggestion
Like with the integrations object above, you'll need to use a plugin to access and modify the context object. For example, if you'd like to add `context.groupId` to every Track call, you should create a plugin like this:
```

forstisabella (approved):

PR #4867, opened 2023-06-12
TikTok Audiences Doc Update with createAudience

rvadera12:
### Proposed changes

We added a new feature where customers can createAudiences through Segment.


rchinn-segment (changes requested):

rchinn-segment (approved):
Thanks for submitting, @rvadera12. We'll include this in our docs deploy tomorrow.

PR #4878, opened 2023-06-14
Add to Engage FAQs

Panandhan22:
## Why the user count in a journey step is greater than the entry/previous step of the journey?

Each step of a Journey is a Personas audience under the hood. The conditions 'stack' so a user has to be a member of the previous step (audience) and meet all the conditions of that current step to get added. Additionally, users cannot exit journeys once they have been added, but they can 'leave' a step (audience) if they no longer meet that step's conditions, even though they will still pass down to the next step. So, if the user no longer satisfies the entry condition will leave that step itself and user count will be reduced. And the user entered the next step will remains there until and unless the user fails for that step's condition and the count won't be reduced.


src/engage/faqs.md
```diff
@@ -122,3 +122,7 @@ Yes, Engage supports the ability to send an audience or computed trait to two or
...
 
 If your team would like to avoid receiving the notifications for transient failures, please **[reach out to support](https://segment.com/help/contact/)**, who upon request can disable transient failure notifications.
+
+## Why the user count in a journey step is greater than the entry/previous step of the journey?
+
+Each step of a Journey is a Personas audience under the hood. The conditions 'stack' so a user has to be a member of the previous step (audience) and meet all the conditions of that current step to get added. Additionally, users cannot exit journeys once they have been added, but they can 'leave' a step (audience) if they no longer meet that step's conditions, even though they will still pass down to the next step. So, if the user no longer satisfies the entry condition will leave that step itself and user count will be reduced. And the user entered the next step will remains there until and unless the user fails for that step's condition and the count won't be reduced.
```
rchinn-segment, line 128:
```suggestion
Each step of a Journey is an Engage audience under the hood. The conditions stack, so a user must be a member of the previous step (audience) and meet all conditions to be added to subsequent steps. However, if the user no longer meets entry conditions for a particular step, they'll exit and you'll see the user count reduced. For any subsequent steps a user is still a part of, they'll remain until they no longer meet entry conditions. 
```

rchinn-segment (changes requested):

PR #4888, opened 2023-06-15
Add blurb when source isn't available in catalog

sarahrudy:
### Proposed changes
- Remove extra "Create a source block" (redundant)
- Add paragraph about alternatives when source isn't available in our catalog

### Merge timing
- ASAP once approved


src/connections/sources/index.md
```diff
@@ -48,6 +48,8 @@ analytics.identify('user_123', {
 {% endcodeexample %}
 {% endcomment %}
 
+> info "If you don't see the source you're looking for in our catalog"
+> If a tool is not listed as a supported source in our [catalog](https://segment.com/catalog/){:target='_blank’}, then it is not possible to incorporate the integration within a Segment workspace out-of-the-box. However, as an alternative, you can use the [HTTP API](https://segment.com/docs/connections/sources/catalog/libraries/server/http-api/){:target='_blank’} source to collect data from the tool's API. You can also use [Functions](https://segment.com/docs/connections/functions/){:target='_blank’} to send or receive data from other tools.
```
forstisabella, line 52: I moved around "out-of-the-box" to describe the integration rather than the workspace. 

Also, links that point to pages in the docs should be relative links (which I do by copying the entire URL, like `https://segment.com/docs/connections/functions/`, and then just deleting the `https://segment.com` part). Relative links like that also don't need the `{:target='_blank’}` attribute!

```suggestion
> If a tool is not listed as a supported source in our [catalog](https://segment.com/catalog/){:target='_blank’}, then it is not possible to incorporate the integration out-of-the-box within a Segment workspace. However, as an alternative, you can use the [HTTP API](/docs/connections/sources/catalog/libraries/server/http-api/) source to collect data from the tool's API. You can also use [Functions](/docs/connections/functions/) to send or receive data from other tools.
```

forstisabella (changes requested):

src/connections/sources/index.md
```diff
@@ -182,16 +184,3 @@ Reverse ETL sources are data warehouses that enable you to use [Reverse ETL](/do
 Reverse ETL supports [these sources] and Segment is actively working on adding more. If you'd like to request Segment to add a particular source, please note it on the [feedback form](https://airtable.com/shriQgvkRpBCDN955){:target="_blank"}.
 
 
-## Create a source
```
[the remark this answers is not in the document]
sarahrudy: Hi @forstisabella, yes, this is an older PR where I made multiple changes, but you're correct! The duplicate "create a source" block has been removed already!

src/connections/sources/index.md
sarahrudy: Ok, great! Thank you for letting me know! I'll keep this in mind moving forward.

forstisabella (approved):

PR #4900, opened 2023-06-20
Showing higher audience count in destination than the Segment UI.

Panandhan22:
## How can be the audience count is high in the connected destination than the Segment UI?

This can happen only when the audiences with same name was created earlier and connected to the corresponding destination. And later if the audience was deleted and recreated with the same name is connected to the same destination. Users of the audience created previously was still available on the destination, then added to that, the audience newly created will also send the users to the destination. Hence, the audience count will be high in destination than the Segment UI.


src/engage/faqs.md
```diff
@@ -36,7 +36,7 @@ You can't change the audience key after it's created. To change the key, you nee
 
 ## Can I reuse audience keys?
 
-Avoid using the same audience key twice, even if you've deleted the key's original audience. Downstream tools and Destinations might have trouble distinguishing between different audiences that at any point shared the same key.
+Avoid using the same audience key twice, even if you've deleted the key's original audience. Downstream tools and Destinations might have trouble distinguishing between different audiences that at any point shared the same key. And this may create mismatch on the audience size between Segment and Destination as the destination may have more audience count as the users of the old audience still exist over the destination.
```
markzegarelli, line 39:
```suggestion
Avoid using the same audience key twice, even if you've deleted the key's original audience. Downstream tools and destinations might have trouble distinguishing between different audiences that once shared the same key. This may create mismatch in audience size between Segment and the destination because the destination may count users of the old audience, resulting in a larger audience size.
```

markzegarelli (approved):

PR #4925, opened 2023-06-22
Add info on track event for openURL method

niallzato:
### Proposed changes
Add info on track event that is triggered from the openURL method

### DIFF
--- a/src/connections/sources/catalog/libraries/mobile/apple/implementation.md
+++ b/src/connections/sources/catalog/libraries/mobile/apple/implementation.md
@@ -331,7 +331,7 @@ var body: some Scene {
     }
 }
 ```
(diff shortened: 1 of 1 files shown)


src/connections/sources/catalog/libraries/mobile/apple/implementation.md
```diff
@@ -331,7 +331,7 @@ var body: some Scene {
     }
 }
 ```
-
+Calling this method with a valid URL parameter, will also trigger a Segment `Deep Link Opened` track event.
```
stayseesong, line 334:
```suggestion
If you call this method with a valid URL parameter, a Segment `Deep Link Opened` track event triggers. 
```

stayseesong (changes requested):

stayseesong (approved):

PR #4929, opened 2023-06-22
Add note on mapping limitation

bobbyatsegment:
### Proposed changes

Track events can only be mapped to a single custom object - even if you try updating the settings to do so, the UI will only save the last mapping for a given event. 

Added note to docs that highlight this and clarify that mapping a single track event to multiple custom objects is not supported.

### DIFF
--- a/src/connections/destinations/catalog/eloqua/index.md
+++ b/src/connections/destinations/catalog/eloqua/index.md
@@ -124,7 +124,7 @@ in Eloqua and a property called `AccountType` in your Segment event, the
 mapping will get handled.
 
 > info "Track event mapping limitations"
-> Each track event can only be mapped to a single custom object; mapping a single track event to multiple custom objects is not supported.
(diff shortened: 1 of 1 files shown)


src/connections/destinations/catalog/eloqua/index.md
```diff
@@ -123,6 +123,9 @@ formatting-insensitive match so that if you have a field called `Account Type`
 in Eloqua and a property called `AccountType` in your Segment event, the
 mapping will get handled.
 
+> info "Track event mapping limitations"
+> Each track event can only be mapped to a single custom object; mapping a single track event to multiple custom objects is not supported.
```
forstisabella, line 127:
```suggestion
> Each `track` event can only be mapped to a single custom object. Segment doesn't support mapping a single `track` event to multiple custom objects.
```

forstisabella (approved):

PR #5031, opened 2023-07-17
include Function authorization example

spencerattick:
### Proposed changes

A customer was confused about how to authenticate the Profile API through Segment Functions so I added an example for future reference.

### Merge timing
ASAP is fine

### DIFF
--- a/src/unify/profile-api.md
+++ b/src/unify/profile-api.md
@@ -169,7 +169,7 @@ When you make requests to the Profile API, use the Access Token as the basic aut
 curl https://profiles.segment.com/v1/spaces/<space_id>/collections/users/profiles
   -u $SEGMENT_ACCESS_TOKEN:
 ```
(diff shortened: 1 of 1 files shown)


src/unify/profile-api.md
```diff
@@ -169,7 +169,15 @@ When you make requests to the Profile API, use the Access Token as the basic aut
 curl https://profiles.segment.com/v1/spaces/<space_id>/collections/users/profiles
   -u $SEGMENT_ACCESS_TOKEN:
 ```
+If you're using a Segment Function or Node.js you can format your header object to include authentication like so:
```
rchinn-segment, line 172:
```suggestion
If you're using a Segment Function or Node.js you can format your header object to include authentication, like so:
```

rchinn-segment (approved):

PR #5038, opened 2023-07-19
Include dot notation limitation

senabon:
Actions Mappings don't support dot notation in mapping fields

### Proposed changes

Several customers are sending in payloads with dot notation and then trying to map them but are unsuccessful because dot notation isn't supported. 

### Merge timing
- ASAP once approved

### Related issues (optional)

https://segment.zendesk.com/agent/tickets/516943 
https://twilio.slack.com/archives/CC97A542H/p1681243270312659

### DIFF
--- a/src/connections/destinations/actions.md
+++ b/src/connections/destinations/actions.md
@@ -174,7 +174,7 @@ You can combine criteria in a single group using **ALL** or **ANY**.  Use an ANY
 > info "Unsupported Special Characters"
 > Mappings do not support the use of double quotes " or a tilde ~ in the trigger fields.
 
->info "Limitations"
+> info "Limitations"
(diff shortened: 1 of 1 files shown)


src/connections/destinations/actions.md
```diff
@@ -174,6 +174,9 @@ You can combine criteria in a single group using **ALL** or **ANY**.  Use an ANY
 > info "Unsupported Special Characters"
 > Mappings do not support the use of double quotes " or a tilde ~ in the trigger fields.
 
+>info "Limitations"
+>Mapping fields do not support dot notation. i.e. properties.amount.cost or properties_amount.cost.
```
stayseesong, line 178:
```suggestion
> Mapping fields don't support dot notation. For example, properties.amount.cost or properties_amount.cost aren't supported.
```

stayseesong (changes requested):

PR #5041, opened 2023-07-19
Add Client ID snippet to Gtag code example

bobbyatsegment:
### Proposed changes

On sending events from both browser and server, [GA4 Web docs mention](https://segment.com/docs/connections/destinations/catalog/actions-google-analytics-4-web/#send-events-from-both-the-browser-and-the-server):

_With Google Analytics 4 Web, events are sent from the browser to GA4. If you use Segment’s [Google Analytics 4 Cloud destination](https://segment.com/docs/connections/destinations/catalog/actions-google-analytics-4/#benefits-of-google-analytics-4-cloud) to send events through the API and tie data between client-side and server-side, you need to pass the same Client ID from the browser and the server. To do this, fetch the Gtag-generated clientId on the web client and pass it to Segment as a property. For more information, see [Google Analytics 4 destination: User Identification](https://segment.com/docs/connections/destinations/catalog/actions-google-analytics-4/#user-identification) on the next steps._

However, the GA4 Cloud Mode docs' section on using Gtag.js and GA4 Cloud mode do not show an example of retrieving the Client ID with Gtag, nor does it directly state that the user identification between the two implementations should be the same. 

Modified the code example to demonstrate this, along with directly stating:
_Additionally, client_id must also align between client-side and server-side events in order to deduplicate user counts in GA4 (unless User-ID is used as the basis for user identification)._


src/connections/destinations/catalog/actions-google-analytics-4/index.md
```diff
@@ -94,9 +94,9 @@ Besides Engagement Time in Milliseconds, you can also generate your own `session
...
 
-When using Gtag, [Google generates a `session_id` and `session_number` when a session begins](https://support.google.com/analytics/answer/9191807?hl=en){:target='_blank'}. The `session_id` and `session_number` generated on the client can be passed as Event Parameters to stitch events sent through the API with the same session that was collected client-side. For events to stitch properly, server-side events must arrive within a 48 hour window of when the client-side events arrive.
+When using Gtag, [Google generates a `session_id` and `session_number` when a session begins](https://support.google.com/analytics/answer/9191807?hl=en){:target='_blank'}. The `session_id` and `session_number` generated on the client can be passed as Event Parameters to stitch events sent through the API with the same session that was collected client-side. Additionally, `client_id` must also align between client-side and server-side events in order to deduplicate user counts in GA4 (unless User-ID is used as the basis for user identification). For events to stitch properly, server-side events must arrive within a 48 hour window of when the client-side events arrive.
```
forstisabella, line 97:
```suggestion
When using Gtag, [Google generates a `session_id` and `session_number` when a session begins](https://support.google.com/analytics/answer/9191807?hl=en){:target='_blank'}. The `session_id` and `session_number` generated on the client can be passed as Event Parameters to stitch events sent through the API with the same session that was collected client-side. Additionally, `client_id` must be the same for both client-side and server-side events in order to deduplicate user counts in GA4 (unless User-ID is used as the basis for user identification). For events to stitch properly, server-side events must arrive within a 48 hour window of when the client-side events arrive.
```

forstisabella (approved):

PR #5051, opened 2023-07-20
Add Listrak Destination Documentation

colevohs:
### Proposed changes

We added documentation for our new Listrak actions destination that was recently put into private beta status.

### Merge timing

This should be published ASAP, but only displayed in the catalog listing once the Listrak actions destination is in public beta status.


stayseesong:
@colevohs I'm going to close this PR out as the docs team is unable to edit it. I've created an editable version here at #5060 that you can follow along with and edit or add comments.

PR #5065, opened 2023-07-25
Add info on canonical url

niallzato:
### Proposed changes
Add info on how the canonical element interacts with the page event

### DIFF
--- a/src/connections/sources/catalog/libraries/website/javascript/index.md
+++ b/src/connections/sources/catalog/libraries/website/javascript/index.md
@@ -270,7 +270,7 @@ analytics.page('Pricing', {
 });
 ```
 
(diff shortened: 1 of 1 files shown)


src/connections/sources/catalog/libraries/website/javascript/index.md
```diff
@@ -270,6 +270,8 @@ analytics.page('Pricing', {
 });
 ```
 
+The `path` and `url` property will be set to the value of the canoncial element on your page. If a canonical element is not set, the values will be set from the browser.
```
rchinn-segment, line 273:
```suggestion
Segment sets the `path` and `url` property to the value of the canonical element on your page. If a canonical element is not set, the values will be set from the browser. 
```

rchinn-segment (approved):

PR #5204, opened 2023-08-16
add Insert Function

spencerattick:
### Proposed changes

Insert Functions also allow folks to filter/alter their data so I wanted to make sure that was noted here.

### Merge timing
ASAP is fine

### DIFF
--- a/src/guides/filtering-data.md
+++ b/src/guides/filtering-data.md
@@ -108,6 +108,10 @@ If you have Protocols in your workspace, **and** have a tracking plan associated
 ![Schema Configuration section of a source's Settings page](images/protocols-unplanned.png)
 
 
+## Destination Insert Function
+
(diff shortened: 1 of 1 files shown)


src/guides/filtering-data.md
```diff
@@ -108,6 +108,10 @@ If you have Protocols in your workspace, **and** have a tracking plan associated
 ![Schema Configuration section of a source's Settings page](images/protocols-unplanned.png)
 
 
+## Destination Insert Function
+
+The newest, most customizable way to filter or alter data going from a source to a Cloud Mode destination is to use [Insert Functions]([url](https://segment.com/docs/connections/functions/insert-functions/)). This feature gives you the abiltiy to receive data from your Segment source, write custom code to alter or block it, and then pass that altered payload to a downstream Cloud Mode destination.
```
stayseesong, line 113:
```suggestion
A customizable way to filter or alter data going from a source to a cloud-mode destination is to use [Insert Functions](/docs/connections/functions/insert-functions/)). This feature gives you the ability to receive data from your Segment source, write custom code to alter or block it, and then pass that altered payload to a downstream cloud-mode destination.
```

stayseesong (changes requested):

PR #5221, opened 2023-08-18
Add array info for CSharp

niallzato:
### Proposed changes
Add array info for CSharp


stayseesong:
@niallzato I made some edits, lmk what you think

stayseesong (approved):

PR #5238, opened 2023-08-23
Update index.md

Hopes-Chantel:
Added IP allowlisting section

### DIFF
--- a/src/connections/destinations/index.md
+++ b/src/connections/destinations/index.md
@@ -172,3 +172,14 @@ You can see the current destination endpoint API success rates and final deliver
 
 > warning ""
 > If you submitted [`suppress_only` requests](https://segment.com/docs/privacy/user-deletion-and-suppression/#suppressed-users), Segment still retains historical events for those users, which can be replayed. If you do not want historical events replayed for suppressed users, submit `suppress_and_delete` requests instead.
+
+### IP Allowlisting
+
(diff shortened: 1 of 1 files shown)


src/connections/destinations/index.md
```diff
@@ -172,3 +172,14 @@ You can see the current destination endpoint API success rates and final deliver
...
+
+### IP Allowlisting
+
+In terms of data leaving Segment’s servers to go to various destinations (not including warehouses), Segment uses Amazon Web Services (AWS), utilizing many different machines in order to send requests. 
+
+These IP addresses can be found [here](https://ip-ranges.amazonaws.com/ip-ranges.json). If you’d like to do so, you will need to allowlist all the IPs from your workspace’s location range.
+
+- US Workspace— AWS us-west-2
+
+- EU workspace— AWS eu-west-1
```
stayseesong, line 184:
```suggestion
When data leaves Segment's servers to go to various destinations (not including warehouses), Segment uses Amazon Web Services (AWS) and utilizes many different machines in order to send requests. 

The IP addresses that are used to send these requests can be found [here](https://ip-ranges.amazonaws.com/ip-ranges.json){:target="_blank"}. If you want to allowlist these specific IP addresses, you need to allowlist all of the IP addresses from your workspace's location range. Below are the ranges: 
* For a US workspace: `AWS us-west-2`
* For an EU workspace: `AWS eu-west-1 `
```

stayseesong (changes requested):

PR #5296, opened 2023-09-05
explain the dedupe window

terence1988:
It’s more concise to publish that we do not only deduplicating `meesageId` within a 24-hour rolling window

### DIFF
--- a/src/guides/duplicate-data.md
+++ b/src/guides/duplicate-data.md
@@ -6,7 +6,7 @@ Segment guarantees that 99% of your data won't have duplicates within an approxi
 
 ## 99% deduplication
 
-Segment has a special deduplication service that sits behind the `api.segment.com` endpoint and attempts to drop 99% of duplicate data. To achieve it, Segment stores at least 24 hours worth of event `messageId`s, which allows Segment systems to deduplicate any data that appears with the same `messageId` within the stored values.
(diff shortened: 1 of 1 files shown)


src/guides/duplicate-data.md
```diff
@@ -6,7 +6,7 @@ Segment guarantees that 99% of your data won't have duplicates within an approxi
 
 ## 99% deduplication
 
-Segment has a special deduplication service that sits behind the `api.segment.com` endpoint and attempts to drop 99% of duplicate data. Segment stores at least 24 hours worth of event `messageId`s, allowing Segment to deduplicate any data that appears within a 24 hour rolling window.
+Segment has a special deduplication service that sits behind the `api.segment.com` endpoint and attempts to drop 99% of duplicate data. To achieve it, Segment stores at least 24 hours worth of event `messageId`s, which allows Segment systems to deduplicate any data that appears with the same `messageId` within the stored values.
```
stayseesong, line 9:
```suggestion
Segment has a special deduplication service that sits behind the `api.segment.com` endpoint and attempts to drop 99% of duplicate data. Segment stores at least 24 hours' worth of event `messageId`s, which allows Segment to deduplicate any data that appears with the same `messageId` within the stored values.
```

stayseesong (changes requested):

stayseesong (approved):

PR #5311, opened 2023-09-07
Add note on delivery overview if no mapping for event

bobbyatsegment:
### Proposed changes

Added note to FAQs that clarifies that a destination's Delivery Overview can show an event as being discarded due to "no matching mappings" if the destination does not have a mapping enabled and configured to handle it.

### DIFF
--- a/src/connections/destinations/actions.md
+++ b/src/connections/destinations/actions.md
@@ -194,7 +194,7 @@ When you send an event with an actions destination Event Tester that doesn't mat
 If no mappings are enabled to trigger on an event that has been received from the connected source, the destination will not send any events. Ensure that at least one mapping has been configured and enabled in the destination mappings for an event that you would like to reach downstream. 
 
 > info ""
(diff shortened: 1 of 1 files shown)


src/connections/destinations/actions.md
```diff
@@ -193,6 +193,9 @@ When you send an event with an actions destination Event Tester that doesn't mat
 
 If no mappings are enabled to trigger on an event that has been received from the connected source, the destination will not send any events. Ensure that at least one mapping has been configured and enabled in the destination mappings for an event that you would like to reach downstream. 
 
+> info ""
+> Events without mappings enabled to handle them can show as being discarded due to "No matching mapping" in a destination's Delivery Overview.
```
rchinn-segment, line 197:
```suggestion
> Events without mappings enabled to handle them display as being discarded due to "No matching mapping" in a destination's Delivery Overview.
```

rchinn-segment: @bobbyatsegment just to confirm, does this behavior occur every time?

rchinn-segment (changes requested):

PR #5362, opened 2023-09-18
Update privacy-portal.md with synonyms example

muhammadatifjav:
### Proposed changes

Added an example of synonyms as customers get confused on how to use this functionality works until you explain their specific use case. 

### Merge timing
- ASAP once approved?

### Related issues (optional)
https://segment.atlassian.net/browse/KCS-626


src/privacy/portal.md
```diff
@@ -223,6 +223,21 @@ North American phone number (without country code) is also ten digits.
 
 {% include components/reference-button.html href="/segment-app/iam/roles/#pii-access" icon="media/academy.svg" title="PII Access" description="Use Access Roles to control who has access to the PII identified by your matchers." variant="related" %}
 
+### Using Synonyms
+
+The matchers scan data coming from your Sources for PII based on both
+exact-matching (and fuzzy-matching if you select 'Similar to' option).
+As an example, if you have a custom matcher for detecting "credit card
+number", it will also detect variations like "credit-card number", or
+typos such as "creditt card number".
```
forstisabella, line 232: Hi @muhammadatifjav - given that we discuss matching in detail in the section immediately before this, I think we can remove this section. 
```suggestion
```

src/privacy/portal.md
```diff
@@ -223,6 +223,21 @@ North American phone number (without country code) is also ten digits.
...
+
+Our fuzzy matching uses [this public GitHub repository](https://github.com/imjasonmiller/godice)
+to determine if it is a match.  However, it will not detect all variations
+and for those scenarios, you can use synonyms. You can add "credit card no",
+"debit card number", or similar variations in the synonyms section.
```
forstisabella, line 237:
```suggestion
Segment's exact matching and fuzzy matching do not detect all variations in the received keys
and for those scenarios, you can use synonyms.  For example, for the value `credit card number`, you can add "credit card no",
"debit card number", or similar variations in the synonyms section to classify those fields. 
```

forstisabella (approved):

PR #5363, opened 2023-09-18
Add Group ID is for mapping of track event to group key $group_id

tanjinhong72:
### Proposed changes
In the Troubleshooting section, modify the following (original) statement:

"Ensure that the mapping(s) handling your track events have the field for Group ID mapped to a valid value. By default, this field is mapped to the event variable context.groupId."

To the following that mentioned using of group key $group_id:

"For Mixpanel (Actions) destination that uses $group_id as the group key, ensure that the mapping(s) handling your track events have the field for Group ID mapped to a valid value. By default, this field is mapped to the event variable"context.groupId."

Customer included a valid value in Group ID but still did not see the track event mapped to the group in Mixpanel. After customer created the group under the group key $group_id, then the track event was mapped to the relevant group.

### Related issues (optional)

https://segment.zendesk.com/agent/tickets/520524
https://segment.zendesk.com/agent/tickets/518710

### DIFF
--- a/src/connections/destinations/catalog/actions-mixpanel/index.md
+++ b/src/connections/destinations/catalog/actions-mixpanel/index.md
@@ -133,5 +133,5 @@ If you want to confirm, you can configure the new destination to point to a diff
 
 ### Track events are not attributed to Mixpanel Groups
 
-Ensure that the mapping(s) handling your `track` events have the field for **Group ID** mapped to a valid value. By default, this field is mapped to the event variable `context.groupId`.
+For Mixpanel (Actions) destination that uses $group_id as the group key, ensure that the mapping(s) handling your `track` events have the field for **Group ID** mapped to a valid value. By default, this field is mapped to the event variable `context.groupId`.
 


src/connections/destinations/catalog/actions-mixpanel/index.md
```diff
@@ -133,5 +133,5 @@ If you want to confirm, you can configure the new destination to point to a diff
 
 ### Track events are not attributed to Mixpanel Groups
 
-Ensure that the mapping(s) handling your `track` events have the field for **Group ID** mapped to a valid value. By default, this field is mapped to the event variable `context.groupId`.
+For Mixpanel (Actions) destination that uses $group_id as the group key, ensure that the mapping(s) handling your `track` events have the field for **Group ID** mapped to a valid value. By default, this field is mapped to the event variable `context.groupId`.
```
stayseesong, line 136:
```suggestion
If the Mixpanel (Actions) destination uses $group_id as the group key, ensure that the mappings handling your `track` events have the field for **Group ID** mapped to a valid value. By default, this field maps to the event variable `context.groupId`.
```

stayseesong (changes requested):

stayseesong (approved):

PR #5401, opened 2023-09-21
Destination Insert Functions GA updates

rchinn-segment:
### Proposed changes

Added Destination Insert Functions GA updates
### Merge timing

9/25 @ 10:00am pst
### Related issues (optional)

DOC-732


src/connections/functions/insert-functions.md
```diff
@@ -237,21 +234,172 @@ To enable your insert function:
...
+> info ""
+> Batching is available for destination and destination insert functions only. 
+
+### When to use batching
+
+Consider creating a batch handler if:
+
+- **You have a high-throughput function and want to reduce cost.** When you define a batch handler, Segment invokes the function once per *batch*, rather than once per event. As long as the function’s execution time isn’t adversely affected, the reduction in invocations should lead to a reduction in cost.
+
+- **Your destination supports batching**. When your downstream destination supports sending data downstream in batches you can define a batch handler to avoid throttling. Batching for functions is independent of batch size supported by the destination. Segment automatically handles batch formation for destinations internally.
```
forstisabella, line 250: Does internally refer to the batch formation or the location of the destination?

```suggestion
- **Your destination supports batching**. When your downstream destination supports sending data downstream in batches you can define a batch handler to avoid throttling. Batching for functions is independent of batch size supported by the destination. Segment automatically handles batch formation for destinations.
```

forstisabella (approved):

src/connections/functions/insert-functions.md
rchinn-segment: Good question, let me confirm with the functions team. I thought it was batch formation, but either way I agree with the suggestion to remove that word completely.

PR #5421, opened 2023-09-26
Add section on Set Once/Set Always fields

bobbyatsegment:
### Proposed changes

[Amplitude's docs](https://www.docs.developers.amplitude.com/analytics/apis/identify-api/#user_properties-supported-operations) mention the following on user_properties operations:

"You can't mix user property operations with actual top-level user properties. Instead, include them inside the $set operation. If you are using one of these operators then this dictionary can contain only user property operations and you can't combine it with the above format. For example. you can't do {"$append":{"interests":"Music"}, "subscription type":"paid"} in the same request."

Add section to docs that calls out this limitation and how to circumnavigate it when using this destination.

### DIFF
--- a/src/connections/destinations/catalog/actions-amplitude/index.md
+++ b/src/connections/destinations/catalog/actions-amplitude/index.md
@@ -216,7 +216,15 @@ Keep the following in mind if you plan to move to Amplitude (Actions) from a cla
 You configure the Amplitude (Actions) destination through Filters and Actions. Consult the table below for information about configuring your Amplitude (Actions) destination similarly to your classic Amplitude destination.
 
 > info ""
-> Contact Segment support if you find features missing from the Amplitude (Actions) destination that were available in the classic Amplitude destination.
+> Contact Segment support if you find features missing from the Amplitude (Actions) destination that were available in the classic Amplitude destination. 
+
+### Set Once/Set Always fields
+
+Amplitude restricts the mixing of top-level user properties with $set, $setOnce, or $setAlways operations in a single request, [as outlined in Amplitude’s documentation](https://www.docs.developers.amplitude.com/analytics/apis/identify-api/#user_properties-supported-operations). 
+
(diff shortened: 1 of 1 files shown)


src/connections/destinations/catalog/actions-amplitude/index.md
```diff
@@ -216,7 +216,15 @@ Keep the following in mind if you plan to move to Amplitude (Actions) from a cla
...
 
 > info ""
-> Contact Segment support if you find features missing from the Amplitude (Actions) destination that were available in the classic Amplitude destination.
+> Contact Segment support if you find features missing from the Amplitude (Actions) destination that were available in the classic Amplitude destination. 
+
+### Set Once/Set Always fields
+
+Amplitude restricts the mixing of top-level user properties with $set, $setOnce, or $setAlways operations in a single request, [as outlined in Amplitude’s documentation](https://www.docs.developers.amplitude.com/analytics/apis/identify-api/#user_properties-supported-operations). 
+
+To circumvent this within Segment, users can opt to exclusively map event parameters to either the User Properties field or to one of the user property operations (Set Once and/or Set Always) available in mappings, avoiding mixing them in the same request. Specifically, if you are utilizing Set Once and/or Set Always fields, include all relevant fields in their respective mappings and do not configure mappings for User Properties in the same request.
```
rchinn-segment, line 225:
```suggestion
To circumvent this within Segment, users can opt to exclusively map event parameters to either the User Properties field or to one of the user property operations (Set Once and/or Set Always) available in mappings. If you are utilizing Set Once and/or Set Always fields, include all relevant fields in their respective mappings and do not configure mappings for User Properties in the same request.
```

src/connections/destinations/catalog/actions-amplitude/index.md
```diff
@@ -216,7 +216,15 @@ Keep the following in mind if you plan to move to Amplitude (Actions) from a cla
...
+
+Conversely, to send top-level user properties, map only to the User Properties field and exclude mappings for Set Once and Set Always fields.
```
rchinn-segment, line 227:
```suggestion
Conversely, to send top-level user properties, map only to the User Properties field and exclude mappings for the Set Once and Set Always fields. 
```

rchinn-segment (changes requested):

rchinn-segment (approved):

PR #5449, opened 2023-09-29
Added information on Journey randomized splits

olesia-ag:
### Proposed changes

Added information on Journey randomized splits - if the Journey has a re-entry condition, users will be joining the same branches of the randomized split as they joined during the first entry.

### Merge timing

- ASAP once approved


src/engage/journeys/step-types.md
```diff
@@ -40,6 +40,9 @@ Define the number of branches you want to create, then add an **Add a condition*
 
 A randomized split lets you experiment with and test the performance of a Journey's branches. When you create a randomized split, you add up to five Journey branches, each with a different step. Journeys then sends eligible users down one of the branches at random. Each branch receives a portion of the eligible users based on percentages that you assign to the branches.
 
+> info ""
+> If the Journey has a re-entry condition, users will be joining the same split branches upon re-entry.
```
rchinn-segment, line 44:
```suggestion
> If the Journey has a re-entry condition, users will join the same split branches upon re-entry.
```

rchinn-segment (approved):

PR #5489, opened 2023-10-05
Add note on custom traits for deleted audiences

bobbyatsegment:
### Proposed changes

Docs do not currently mention how an audience key can show up as a custom trait on an Engage profile after the corresponding audience is deleted (provided the audience previously generated identify calls for users) - added as a note under "Custom Traits" section.

### DIFF
--- a/src/engage/audiences/index.md
+++ b/src/engage/audiences/index.md
@@ -35,6 +35,9 @@ You can also specify two different types of time-windows, `within` and `in betwe
 
 You can also build Audiences based on custom traits. These traits can be collected from your apps when a user completes a form or signs up using an [Identify](/docs/connections/spec/identify) call. You can view these traits in the Profile explorer, as well. Custom Traits are mutable and update to the latest value seen by the user's Identify events.
 
+> info ""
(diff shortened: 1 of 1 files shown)


src/engage/audiences/index.md
```diff
@@ -35,6 +35,9 @@ You can also specify two different types of time-windows, `within` and `in betwe
 
 You can also build Audiences based on custom traits. These traits can be collected from your apps when a user completes a form or signs up using an [Identify](/docs/connections/spec/identify) call. You can view these traits in the Profile explorer, as well. Custom Traits are mutable and update to the latest value seen by the user's Identify events.
 
+> info ""
+> When an audience that previously generated Identify events is deleted, the data for the audience key is still attached to profiles that entered the audience, and becomes visible in the UI as a custom trait.
```
rchinn-segment, line 39:
```suggestion
> When an audience that previously generated Identify events is deleted, the data for the audience key is still attached to profiles that entered the audience, and becomes visible in Segment as a custom trait.
```

rchinn-segment (approved):

PR #5551, opened 2023-10-18
Update hipaa-eligible-segment.md

sneha-shashidhar:
### Proposed changes

Customers with HIPAA eligible workspaces often write to CSEng about why they're being logged out of their workspaces frequently. This is rather expected given that a factor of becoming HIPAA compliant is that the session logs out after 15 minutes of inactivity. It'd be great to highlight this in our Public Docs.


forstisabella:
let's go ahead and close this - the security team opened #5570 which addresses this concern.

PR #5575, opened 2023-10-24
Source transfer impact on connected destination

Panandhan22:
### Proposed changes

> info ""
> When a source is transferred from one workspace to another, it will include all the connected destinations with the same settings and nothing will change from the source details. The only difference is that the source will now belongs to a different workspace.


forstisabella (approved):

src/guides/usage-and-billing/account-management.md
```diff
@@ -49,6 +49,8 @@ Though workspaces can't be merged, you can move an existing source to a single w
 
 To move a source between workspaces, navigate to the source's **Settings** tab, then click **Transfer to Workspace**. Choose the workspace you're moving the source to, then click **Transfer Source**.
 
+When a source is transferred from one workspace to another, it will include all the connected destinations with the same settings.
```
forstisabella, line 52:
```suggestion
When you transfer a source from one workspace to another, the new source includes all the connected destinations with the same settings.
```

PR #5577, opened 2023-10-24
Fix Duplicate Table of Contents in `/connections/spec/mobile`

joshpaulchan:
Remove repeated table of contents in mobile spec

### Proposed changes

Saw there were duplicate table of contents in https://segment.com/docs/connections/spec/mobile/ and figured I'd fix it.


forstisabella (approved):
lgtm! thanks for catching this

PR #5604, opened 2023-10-31
Use account-level SQL traits to associate users to the account

Panandhan22:
### Use account-level SQL traits to associate users to the account

To associate users to an account via SQL traits, we have to return both group_id and user_id in the account level SQL trait - this will fire a group call with the userId and the groupId in question, which is the method that Segment uses to add users to groups in destinations.

There is a limitation on Segment end when a group (account) contains more than one user, the query will return duplicate group_ids (mapped to unique user_ids). Segment won't allow us to return duplicate group_id's in the Group(account level) SQL trait., So we can't map the users to the accounts on many-to-many situation.

### DIFF
--- a/src/engage/audiences/account-audiences.md
+++ b/src/engage/audiences/account-audiences.md
@@ -66,7 +66,7 @@ Use-cases for account-level computed traits include:
 > info ""
 > Use SQL traits for complex calculations not supported by computed traits. For example, you would use SQL traits to calculate the number of unique users associated with an account who have logged in during the past month.
 
-### Use account-level SQL traits to associate users to the account
+### Use account-level SQL traits to associate users to an account
 
 To associate users to an account via SQL traits, we have to return both group_id and user_id in the account level SQL trait - this will fire a group call with the userId and the groupId in question, which is the method that Segment uses to add users to groups in destinations.
 
--- a/src/engage/audiences/account-audiences.md
+++ b/src/engage/audiences/account-audiences.md
@@ -70,7 +70,7 @@ Use-cases for account-level computed traits include:
 
 To associate users to an account with SQL traits, you must return both the `group_id` and `user_id` in the account level SQL trait. This fires a Group call which Segment uses to add users to groups in destinations.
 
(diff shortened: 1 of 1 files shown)


src/engage/audiences/account-audiences.md
```diff
@@ -66,6 +66,12 @@ Use-cases for account-level computed traits include:
 > info ""
 > Use SQL traits for complex calculations not supported by computed traits. For example, you would use SQL traits to calculate the number of unique users associated with an account who have logged in during the past month.
 
+### Use account-level SQL traits to associate users to the account
+
+To associate users to an account via SQL traits, we have to return both group_id and user_id in the account level SQL trait - this will fire a group call with the userId and the groupId in question, which is the method that Segment uses to add users to groups in destinations.
```
rchinn-segment, line 71:
```suggestion
To associate users to an account with SQL traits, you must return both the `group_id` and `user_id` in the account level SQL trait. This fires a Group call which Segment uses to add users to groups in destinations.
```

src/engage/audiences/account-audiences.md
```diff
@@ -66,6 +66,12 @@ Use-cases for account-level computed traits include:
...
+
+There is a limitation on Segment end when a group (account) contains more than one user, the query will return duplicate group_ids (mapped to unique user_ids). Segment won't allow us to return duplicate group_id's in the Group(account level) SQL trait., So we can't map the users to the accounts on many-to-many situation.
```
rchinn-segment, line 73:
```suggestion
When a group (account) contains more than one user, the query returns duplicate `group_id`s (mapped to unique `user_id`s). However, Segment doesn't return duplicate `group_id`s in the group (account level) SQL trait. As a result, you can't map the users to accounts in a many-to-many situation.
```

src/engage/audiences/account-audiences.md
```diff
@@ -66,6 +66,12 @@ Use-cases for account-level computed traits include:
 > info ""
 > Use SQL traits for complex calculations not supported by computed traits. For example, you would use SQL traits to calculate the number of unique users associated with an account who have logged in during the past month.
 
+### Use account-level SQL traits to associate users to the account
```
rchinn-segment, line 69:
```suggestion
### Use account-level SQL traits to associate users to accounts
```

rchinn-segment (changes requested):

rchinn-segment (approved):

PR #5611, opened 2023-10-31
Updated destination-filters limitations.md

MStephen024:
Added to the destination filter limitations with info on events possibly bypassing the filter for missing required fields.

With classic destinations in particular, if an event type that is not supported by the destination is sent, Segment will automatically fail the event and send it to the Event Delivery view. These events will bypass any enabled destination filters. This also includes events that are missing required fields.

### Proposed changes

*(For non—[actions-based destinations](https://segment.com/docs/connections/destinations/actions/#available-actions-based-destinations)) If the connected source sends an event that is missing a required field, it's possible for the event to bypass any enabled filter and appear in the [event delivery](https://segment.com/docs/connections/event-delivery/) view.


cmastr (approved):

pwseg:
Closing this one out.  It's been updated with #6234.

PR #5625, opened 2023-11-01
docs(RN): update code blocks to ease copy feature

kud:
### Proposed changes

It is better to have two blocks to facilitate the copy button feature.

image

### DIFF
--- a/src/connections/sources/catalog/libraries/mobile/react-native/index.md
+++ b/src/connections/sources/catalog/libraries/mobile/react-native/index.md
@@ -26,14 +26,19 @@ To get started with the Analytics for React Native library:
 
     ```js
     yarn add @segment/analytics-react-native @segment/sovran-react-native react-native-get-random-values
-    # or
+    ```
+
+    ```js
     npm install --save @segment/analytics-react-native @segment/sovran-react-native react-native-get-random-values
     ```
+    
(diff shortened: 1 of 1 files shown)


kud:
Also, one question, is ` @segment/sovran-react-native` still mandatory? It seems that this project is deprecated now.

stayseesong:
@kud thanks for submitting this PR! `@segment/sovran-react-native` isn't mandatory, you can use your own storage solution if you want, but it’s strongly recommended to use it. It's also not deprecated, the repo just moved to here: https://github.com/segmentio/analytics-react-native/tree/master/packages/sovran. I've updated the docs file so that it directs to the correct link.

stayseesong (approved):

kud:
Oh, I see. Thanks!

For some reason, I ended up on the archived GitHub repository. Alright, thanks for the explanation. Is it similar to Zustand?

Regarding the code blocks, I completely understand your point of view regarding the "or" option, which can make sense when reading a README. However, when a user visits your website, the code block is copyable via the button in the top-right corner. If you click on this copy button, you'll get those two instructions, which is not ideal when you want to paste it into your shell. If there are two blocks, you can choose the one you want and directly copy the specific command you need.

https://github.com/segmentio/segment-docs/assets/655838/7614d2e0-e367-47c7-9ef5-4098555af817

PR #5642, opened 2023-11-05
Podscribe destination docs

pitronalldak:
### Proposed changes

Podscribe destination docs

### Merge timing
once approved


rchinn-segment:
@pitronalldak - I'm re-opening this as PR #5668 so that we have edit access

PR #5657, opened 2023-11-09
Added information on timestamp format for time comparison operators

olesia-ag:
### Proposed changes

Added a section on time comparison operators, with a note that they will only work with valid ISO timestamps

### Merge timing

ASAP once approved

### DIFF
--- a/src/engage/audiences/index.md
+++ b/src/engage/audiences/index.md
@@ -42,7 +42,7 @@ You can also use computed traits in an Audience definition. For example, you can
 > info ""
 > Engage supports nested traits, but the Audience builder doesn’t support accessing objects nested in arrays. When you send arrays of objects, they are flattened into strings. As a result, the same conditions that work on strings will work on the array. Within the builder, you can only use string operations like `contains` and `does not contain` to look for individual characters or a set of characters in the flattened array.
 
-### Time Comparision
+### Time comparison
 
 You can use various time comparison operators in your Audience definition. Such operators are: before date, after date, within last, within next, before last, after next. Only ISO timestamps can be used with these operators.
 If the timestamp is not a valid ISO timestamp (for example, a trailing 'Z' is missing), it will not be processed by Real-time compute. See [Real-time compute compared to batch](docs/engage/audiences/#real-time-compute-compared-to-batch)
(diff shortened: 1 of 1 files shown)


src/engage/audiences/index.md
```diff
@@ -42,6 +42,11 @@ You can also use computed traits in an Audience definition. For example, you can
 > info ""
 > Engage supports nested traits, but the Audience builder doesn’t support accessing objects nested in arrays. When you send arrays of objects, they are flattened into strings. As a result, the same conditions that work on strings will work on the array. Within the builder, you can only use string operations like `contains` and `does not contain` to look for individual characters or a set of characters in the flattened array.
 
+### Time Comparision
```
rchinn-segment, line 45:
```suggestion
### Time comparison
```

src/engage/audiences/index.md
```diff
@@ -42,6 +42,11 @@ You can also use computed traits in an Audience definition. For example, you can
...
+
+You can use various time comparison operators in your Audience definition. Such operators are: before date, after date, within last, within next, before last, after next. Only ISO timestamps can be used with these operators.
```
rchinn-segment, line 47:
```suggestion
You can use the following time comparison operators in your audience definition: 
- before date
- after date 
- within last
- within next 
- before last
- after next 

Only ISO timestamps can be used with these operators.
```

src/engage/audiences/index.md
```diff
@@ -42,6 +42,11 @@ You can also use computed traits in an Audience definition. For example, you can
...
+If the timestamp is not a valid ISO timestamp (for example, a trailing 'Z' is missing), it will not be processed by Real-time compute. See [Real-time compute compared to batch](docs/engage/audiences/#real-time-compute-compared-to-batch)
```
rchinn-segment, line 48:
```suggestion
If the timestamp is not a valid ISO timestamp (for example, a trailing `Z` is missing), Segment won't process the audience in real-time. Learn more about [real-time compute compared to batch](docs/engage/audiences/#real-time-compute-compared-to-batch)
```

rchinn-segment (changes requested):

pwseg (approved):

PR #5714, opened 2023-11-21
Spec: Alias rewrite

### LINKED ISSUE #4829: [Content]: Alias doc

https://segment.com/docs/connections/spec/alias/

The alias page doesn't make it clear at all which order the function arguments come in. It maybe sorta implies that it's `alias(previousId, newId)` is the order of arguments, but it isn't demonstrated in an example.

Furthermore, in the TypeScript types, it labels the first argument as `to` and the second argument as `from` but what the heck does that mean? I would assume that `from` is previousId implying that the order of arguments is `alias(newId, previousId)`, but which is it? Can you not just label the arguments `previousId` / `newId` in an example? Or does the order of arguments not matter?

### Additional information

_No response_

### LINKED ISSUE #4940: [Content]: Spec: Alias

https://segment.com/docs/connections/spec/alias/#:~:text=Segment%20University%3A%20The%20Segment%20Methods

It references Segment University but if you look at Segment University, there is no mention of the Alias method (at least that I can find!): 

Screenshot 2023-06-23 at 2 27 38 PM

### Additional information

_No response_

forstisabella:
### Proposed changes
Rewrote Alias doc, added more context from the analytics-js page, Vale fixes

### Merge timing
after approval

### Related issues (optional)
Closes #4940
Closes #4829


src/connections/spec/alias.md
```diff
@@ -2,23 +2,37 @@
 title: 'Spec: Alias'
 ---
 
-> note "Alias is an advanced method"
-> The Alias method allows you to explicitly change the ID of a tracked user, however this should only be done when it's required for downstream destination compatibility. See our [Best Practices for Identifying Users](/docs/guides/how-to-guides/best-practices-identify/) for more information.
+The Alias method is an advanced method used to merge two unassociated user identities, effectively connecting two sets of user data in one profile.
```
rchinn-segment, line 5:
```suggestion
The Alias method is an advanced method used to merge 2 unassociated user identities, effectively connecting 2 sets of user data in one profile. 
```

src/connections/spec/alias.md
```diff
@@ -28,26 +42,19 @@ Here's the payload of a basic `alias` call that will associate this user's exist
...
-
 Here's the corresponding JavaScript event that would generate the above payload. If you're using Segment's JavaScript library, Segment automatically passes in the user's `anonymousId` as `previousId` for you:
 
 ```js
 analytics.alias("507f191e81");
 ```
-{% include content/syntax-note.md %}
 
-Beyond the common fields, the `alias` call takes the following fields:
+If you're instrumenting a website, the Anonymous ID is generated in the browser so you must call Alias from the client-side. If you're using a server-side session ID as the Anonymous ID, then you must Alias from the server-side.
```
rchinn-segment, line 51:
```suggestion
If you're instrumenting a website, the Anonymous ID is generated in the browser so you must call Alias from the client-side. If you're using a server-side session ID as the Anonymous ID, then you must call Alias from the server-side.
```

rchinn-segment (approved):

PR #5733, opened 2023-11-28
Update insert-functions.md

volneyvj:
### Proposed changes

Added a 'tip' for customer who does not need any transformation of a specific event type. 
For example, Insert Functions transforms track events, but there's no need for updates over page() calls. In this case, the handler statement is still necessary, but the user has to at least add a 'return event' inside of it. Otherwise, the event will get dropped.

### DIFF
--- a/src/connections/functions/insert-functions.md
+++ b/src/connections/functions/insert-functions.md
@@ -57,7 +57,8 @@ Segment invokes a separate part of the function (called a "handler") for each ev
 The default source code template includes handlers for all event types. You don't need to implement all of them - just use the ones you need, and skip the ones you don't.
 
 > info ""
-> Removing the handler for a specific event type results in blocking the events of that type from arriving at their destination. 
+> Removing the handler for a specific event type results in blocking the events of that type from arriving at their destination. In case an event type should remain as it is,
+> but you'd still need to parte it dowstream, you may merely add  a 'return event', inside the event type handler statement.
 
 Insert functions can define handlers for each message type in the [Segment spec](/docs/connections/spec/):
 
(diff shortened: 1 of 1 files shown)


src/connections/functions/insert-functions.md
```diff
@@ -57,7 +57,8 @@ Segment invokes a separate part of the function (called a "handler") for each ev
 The default source code template includes handlers for all event types. You don't need to implement all of them - just use the ones you need, and skip the ones you don't.
 
 > info ""
-> Removing the handler for a specific event type results in blocking the events of that type from arriving at their destination. 
+> Removing the handler for a specific event type results in blocking the events of that type from arriving at their destination. In case an event type should remain as it is,
+> but you'd still need to parte it dowstream, you may merely add  a 'return event', inside the event type handler statement.
```
volneyvj, line 61: Removing the handler for a specific event type results in blocking the events of that type from arriving at their destination. In case an event type should not explicitly alter by an Insert Function, but you'd still need to send it downstream, you just need to add an event handler for that event type and include 'return event', inside the event type handler statement.

rchinn-segment, line 61:
```suggestion
```

src/connections/functions/insert-functions.md
```diff
@@ -57,7 +57,8 @@ Segment invokes a separate part of the function (called a "handler") for each ev
 The default source code template includes handlers for all event types. You don't need to implement all of them - just use the ones you need, and skip the ones you don't.
 
 > info ""
-> Removing the handler for a specific event type results in blocking the events of that type from arriving at their destination. 
+> Removing the handler for a specific event type results in blocking the events of that type from arriving at their destination. In case an event type should remain as it is,
```
rchinn-segment, line 60:
```suggestion
> Removing the handler for a specific event type results in blocking the events of that type from arriving at their destination. To keep an event type as is but still send it downstream, add a `return event` inside the event type handler statement.
```

rchinn-segment (approved):

PR #5744, opened 2023-11-30
Add support for location attributes in Attio

nickcampbell18:
### Proposed changes

Hi folks! We've recently added an address parser to the Attio API so we can now accept location data. Since all these attributes are cast down to strings, it accepts a string containing mixed location data and converts it back into structured data internally (it doesn't really matter about ordering of parts, or punctuation, these are stripped out when the parser runs).

Here's an example from the mappings we use for our Segment instance:

Screenshot 2023-11-27 at 09 40 14

Do you think there is any further guidance we should provide here?

### Merge timing

- ASAP once approved

### DIFF
--- a/src/connections/destinations/catalog/actions-attio/index.md
+++ b/src/connections/destinations/catalog/actions-attio/index.md
@@ -158,10 +158,9 @@ with both the domain and twitter handles above.
 
 ## Attribute types
 
-With the exception of location data, the Attio Action can write all other types of
-attribute to Attio. Below is an example of the format that each attribute must be; please
-note that you'll get validation failures if any of these are incorrect. To unset an
-attribute, you can also pass `null` as the value.
+The Attio Action can write all types of attribute to Attio. Below is an example of the
+format that each attribute must be; please note that you'll get validation failures if any
+of these are incorrect. To unset an attribute, you can also pass `null` as the value.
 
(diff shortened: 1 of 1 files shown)


src/connections/destinations/catalog/actions-attio/index.md
```diff
@@ -171,7 +170,7 @@ attribute, you can also pass `null` as the value.
 | `date`               | YYYY-MM-DD                                                                              | `"2023-09-28"`                                              |
 | `domain`             | `{domain}.{tld}`                                                                        | `"app.attio.com"`, `"www.example.com"`                      |
 | `email`              | A valid email address                                                                   | `"person@example.com"`                                      |
-| `location`           | *unsupported*                                                                           |                                                             |
+| `location`           | String with all valid address parts combined                                            | "1 Infinite Loop, Cupertino, CA, US"                        |
```
forstisabella, line 173:
```suggestion
| `location`           | String with all valid address parts (street address, city, state, country, and postal code) combined                                            | "1 Infinite Loop, Cupertino, CA, US"                        |
```

forstisabella: I think this should cover it!

forstisabella (approved):

nickcampbell18:
Thank you!

PR #5765, opened 2023-12-05
Add note on returning event object in handler

bobbyatsegment:
### Proposed changes

Docs don't currently call out that the event object needs to be returned at the end of a method handler in order for the connected destination to proceed with processing the Function-modified event payload. 

Added a note under the section to code the insert function that mentions this.

### DIFF
--- a/src/connections/functions/insert-functions.md
+++ b/src/connections/functions/insert-functions.md
@@ -100,8 +100,7 @@ async function onTrack(event) {
 
 To change which event type the handler listens to, you can rename it to the name of the message type. For example, if you rename this function `onIdentify`, it listens for "Identify" events instead.
 
-> info ""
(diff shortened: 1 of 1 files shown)


src/connections/functions/insert-functions.md
```diff
@@ -100,6 +100,9 @@ async function onTrack(event) {
 
 To change which event type the handler listens to, you can rename it to the name of the message type. For example, if you rename this function `onIdentify`, it listens for "Identify" events instead.
 
+> info ""
+> To ensure the Destination processes an event payload modified by the Function, return the `event` object at the handler's end.
```
stayseesong, line 104:
```suggestion
To ensure the Destination processes an event payload modified by the function, return the `event` object at the handler's end.
```

stayseesong (changes requested):

stayseesong (approved):

PR #5850, opened 2024-01-07
Rename event-delivery.md to event-delivery.md.

feim9208:


rchinn-segment:
Hi @feim9208, why did you want to rename this file?

PR #5886, opened 2024-01-16
Add a note about Chrome third-party cookie restrictions.md

prigiattiperrut:
### Proposed changes

Based on recent support tickets, customers have expressed concerns about Chrome's third-party cookie restrictions and how they will impact Segment and its destinations. Therefore, we are adding a note about this in our documentation so customers can find this information online

- ASAP once approved


src/connections/sources/catalog/libraries/website/javascript/troubleshooting.md
```diff
@@ -179,6 +179,12 @@ Some destinations accept properties only. As a result, custom context fields you
...
+
+Regarding cookies set by [device-mode destinations](https://segment.com/docs/connections/destinations/#connection-modes). It's important to note, that Segment primary function is to load third-party SDKs and forward events to them. As a result, the usage and management of cookies are entirely at the discretion of each individual SDK. For instance, if you have concerns about destinations setting third-party cookies, it would be best to consult directly with the destination providers for detailed information. For example, Amplitude, one of the destinations in our catalog, offers an informative [article on this topic](https://www.docs.developers.amplitude.com/guides/cookies-consent-mgmt-guide/#frequently-asked-questions).
```
stayseesong, line 186:
```suggestion
Regarding cookies set by [device-mode destinations](/docs/connections/destinations/#connection-modes), it's important to note, that Segment's primary function is to load third-party SDKs and forward events to them. As a result, the usage and management of cookies are entirely at the discretion of each individual SDK. For instance, if you have concerns about destinations setting third-party cookies, it's best to consult directly with the destination providers for detailed information. For example, Amplitude, one of the destinations in the Segment catalog, offers an informative [article on this topic](https://www.docs.developers.amplitude.com/guides/cookies-consent-mgmt-guide/#frequently-asked-questions){:target="_blank"}.
```

stayseesong (changes requested):

src/connections/sources/catalog/libraries/website/javascript/troubleshooting.md
```diff
@@ -179,6 +179,12 @@ Some destinations accept properties only. As a result, custom context fields you
 
 The AJS cookies being set under segment.com are first-party cookies. They are part of Segment's own implementation as well as the destination Segment uses. These cookies are not related to your implementation of Segment, and you only see them because you've visited Segment's domain using the same browser. They are sent to the writekey connected to Segment's own workspace, and are associated with the events Segment tracks when you visit segment.com.
 
+## Will Google Chrome's Third-Party Cookie Changes Impact Segment Analytics.js?
+
+The short answer is no. Analytics.js will not be affected by this change. This is because Analytics.js functions as a first-party cookie. Unlike third-party cookies, which are set by external services and blocked by the [new Chrome update](https://developers.google.com/privacy-sandbox/3pcd), Analytics.js is integrated directly into the user's website or application. It sets cookies through the user's own bundled code, thus ensuring its status as a first-party cookie and guaranteeing its operational effectiveness despite Chrome's new policy.
```
chrisradek, line 184: > It sets cookies through the user's own bundled code, thus ensuring its status as a first-party cookie

I think this implies that A.JS isn't being downloaded through cdn.segment.com. It creates a cookie on the domain/sub-domain of the page that it is loaded on and reads this cookie (So if it is loaded on foo.com, the cookies is readable only by foo.com and its subdomains). Whereas 3rd party cookies are created for a different domain than the page it is being read from (e.g. creating a cookie for facebook.com but being able to read it from any website)

stayseesong: @chrisradek should I just take the below part out according to your comment?

> Analytics.js is integrated directly into the user's website or application. It sets cookies through the user's own bundled code, thus ensuring its status as a first-party cookie and guaranteeing its operational effectiveness despite Chrome's new policy.

stayseesong (approved):

PR #5899, opened 2024-01-18
Prediction failure

Panandhan22:
### Proposed changes

> info "Prediction is failing with error "We weren't able to create this prediction because your requested prediction event is not being tracked anymore. Please choose a different prediction event and try again.""
> Predictions are computed based on the available data and the conditions specified for the trait. A gap in tracking events for seven continuous days could potentially affect the computation of the prediction.
Nevertheless, once data tracking resumes and there is enough data, the prediction should be recomputed.

### DIFF
--- a/src/unify/Traits/predictions/index.md
+++ b/src/unify/Traits/predictions/index.md
@@ -99,6 +99,10 @@ Segment would then build the prediction from this criteria and create specific p
 
 Predicted LTV has strict data requirements. Segment can only make predictions for customers that have purchased two or more times. Segment also requires a year of purchase data to perform LTV calculations.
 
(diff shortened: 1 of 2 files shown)


rchinn-segment (commented):
Thanks for submitting, @Panandhan22! Maybe we can add this content to the following section: https://segment.com/docs/unify/traits/predictions/using-predictions/#are-there-any-known-predictions-limitations

cmastr:
> Thanks for submitting, @Panandhan22! Maybe we can add this content to the following section: https://segment.com/docs/unify/traits/predictions/using-predictions/#are-there-any-known-predictions-limitations

Made this change to the location.

cmastr (approved):

pwseg (approved):

PR #5934, opened 2024-01-26
Rvadera12 patch 2

rvadera12:
### Proposed changes

We are deprecating this destination, as we have built out a new DV360 in Actions. This informs customers of the deprecation and migration. 

### Merge timing

On January 30th - when dv360 goes to public beta

### DIFF
--- a/src/connections/destinations/catalog/personas-display-video-360/index.md
+++ b/src/connections/destinations/catalog/personas-display-video-360/index.md
@@ -6,11 +6,14 @@ id: 5d4dd5b989eda01a09b5cdb1
 ---
 
 > warning "Deprecation Notice"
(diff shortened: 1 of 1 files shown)


cmastr (approved):

cmastr:
@rvadera12 Should this documentation be removed from Segment docs on March 6th, 2024 when the deprecation happens? Also if we publish these docs today, is that ok?

rvadera12:
> @rvadera12 Should this documentation be removed from Segment docs on March 6th, 2024 when the deprecation happens? Also if we publish these docs today, is that ok?

@cmastr We should be good to publish today - however if possible could this be published tomorrow. Yes this documentation should be removed from Segment docs on March 6th  (will follow up if there is any change in plans)

cmastr:
@rvadera12  it'll go out today, January 30th. Thanks!

PR #5961, opened 2024-01-31
Add info on schema integration filters with existing integration object

ashton-huxtable:
### Proposed changes

Added information on using per source integration filters when the event payload already has an object value in place for the integration. 

### Merge timing
- ASAP once approved

### Related issues (optional)

n/a

### DIFF
--- a/src/guides/filtering-data.md
+++ b/src/guides/filtering-data.md
@@ -80,6 +80,17 @@ The events filtered out of individual destinations using this method still arriv
 
 **Integration filters are all-or-nothing for each event.** If you require more detailed control over which events are sent to specific destinations, you can use Destination Filters to inspect the event payload, and conditionally drop the data or forward it to the destination.
 
+**Integration filters will not override an existing value in the integrations object.** If the integration object already has a value for the integration, the per source schema integration filters will not override this. For example, if you are sending events to Appsflyer with the appsflyerId passed into the integration object:
+
+```javascript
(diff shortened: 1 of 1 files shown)


src/guides/filtering-data.md
```diff
@@ -80,6 +80,17 @@ The events filtered out of individual destinations using this method still arriv
...
+**Integration filters will not override an existing value in the integrations object.** If the integration object already has a value for the integration, the per source schema integration filters will not override this. For example, if you are sending events to Appsflyer with the appsflyerId passed into the integration object:
+
+```javascript
+integrations: {
+  Appsflyer: {
+    appsflyerId: 'xxxxxx'
+  }
+}
+```
+And for the same event you have Appsflyer turned off using the per source schema integrations filter, this filter will not override the above object with a false value, and events will still send downstream. In this scenario, [destination filters](#destination-filters) can be used to drop the event before it is sent downstream.
```
stayseesong, line 92:
```suggestion
For the same event you have Appsflyer turned off using the per source schema integrations filter, this filter won't override the above object with a false value, and events still send downstream. In this scenario, [destination filters](#destination-filters) can be used to drop the event before sends downstream. 
```

stayseesong (changes requested):

stayseesong (approved):

stayseesong (approved):

PR #5993, opened 2024-02-05
update-warehouse-faq-datatype

terence1988:
### Proposed changes
Adding more details to this FAQ can help customers to understand Segment's limits

### Merge timing
- ASAP once approved


graziellecorrea:
Hey @terence1988 , I'd change the text to the following (as discussed on [Jira](https://segment.atlassian.net/browse/KCS-1075)): 


"Yes. Data types are set up in your warehouse based on the first value that comes in from a source, but you can request data type changes by reaching out to [Segment Support](https://app.segment.com/workspaces?contact=1). Keep in mind that Segment uses [general data types](https://segment.com/docs/connections/storage/warehouses/schema/#data-types) when loading data in your warehouse, for example:

- Changing data type from `timestamp` to `varchar` 

- Changing data type from `integer` to `float`

- Changing data type from `boolean` to `varchar`

More granular changes (such as the examples mentioned below) wouldn’t normally be handled by our Support team. These should be made within the warehouse itself:

- Expanding `varchar(256)` to `varchar(2048)`

- Updating `integer` to `bigint`

- Updating `float` to `float8` "

Thanks!

forstisabella (approved):

PR #6144, opened 2024-02-28
Update Destination Filters device-mode support

PcooperSegment:
### Proposed changes

Adjusting the line for mobile device mode libraries. The older libraries such as IOS and Android don't have support for destination filters. 

Swift and Kotlin do have destination filter support.

### Merge timing

- ASAP once approved


niallzato (changes requested):
Perhaps change to the following:
````
- *(For IOS and Android mobile device-mode)* Destination filters are currently not supported. Our Swift, Kotlin and React Native libraries do support device-mode destination filters.
````

PR #6160, opened 2024-02-29
Add section to address named page events in source ingestion.md

ashton-huxtable:
### Proposed changes

Customers have seen 'unexpected' event names in this section which have turned out to be Page calls that are named. 

### Merge timing
- ASAP once approved

### Related issues (optional)

n/a

### DIFF
--- a/src/connections/sources/index.md
+++ b/src/connections/sources/index.md
@@ -87,7 +87,7 @@ You can use the time picker located on the Source Overview page to specify a tim
 
 The breakdown table displays the following details:
 * **Event type**: The Segment Spec event type (Track call vs. Identify call, for example).
-* **Event name**: The event name, provided by you or the source. You should only expect to see track calls with event names. If you see unnamed events in the breakdown table, it's because for other specs, you're either identifying users/groups or tracking what page/screen the user is on. Both of which do not have event names. The exception to this is if you have provided a name in the Page call. Named page calls will also show up in this section with the name of the Page as the event name. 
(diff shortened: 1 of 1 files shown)


src/connections/sources/index.md
```diff
@@ -87,7 +87,7 @@ You can use the time picker located on the Source Overview page to specify a tim
 
 The breakdown table displays the following details:
 * **Event type**: The Segment Spec event type (Track call vs. Identify call, for example).
-* **Event name**: The event name, provided by you or the source. You should only expect to see track calls with event names. If you see unnamed events in the breakdown table, it's because for other specs, you're either identifying users/groups or tracking what page/screen the user is on. Both of which do not have event names. 
+* **Event name**: The event name, provided by you or the source. You should only expect to see track calls with event names. If you see unnamed events in the breakdown table, it's because for other specs, you're either identifying users/groups or tracking what page/screen the user is on. Both of which do not have event names. The exception to this is if you have provided a name in the Page call. Named page calls will also show up in this section with the name of the Page as the event name.
```
stayseesong, line 90:
```suggestion
* **Event name**: The event name, provided by you or the source. You should only expect to see track calls with event names. If you see unnamed events in the breakdown table, it's because for other specs, you're either identifying users/groups or tracking what page/screen the user is on. Both of which do not have event names. The exception to this is if you provide a name in the Page call. Named Page calls also show up in this section with the name of the Page as the event name. 
```

stayseesong (changes requested):

stayseesong (approved):

PR #6177, opened 2024-03-05
Device mode not supported.md

estgc:
Add note that device mode destinations are not supported as clients still ask this even if it says only cloud mode are

### Merge timing
- ASAP once approved

### Related issues (optional)
https://segment.zendesk.com/agent/tickets/537171


cmastr:
Ryan is no longer at Segment, I'll take this as a review.

PR #6208, opened 2024-03-08
Add section about attribution to GA4 docs

sarahrudy:
### Proposed changes
- Add section about attribution to GA4 docs 
- Include links to Google's documentation so that customers can read more 

### Merge timing
- ASAP once approved


cmastr:
this is live from another PR now.

PR #6244, opened 2024-03-18
Add FAQ for Common Error in HubSpot Cloud Mode

courtneyga:
### Proposed changes

Customers who use HubSpot Cloud Mode and have an “Upsert Contact” mapping and “CBE” mapping, receive a contact already exists error with two other responses, see below:

Response 1:

"category": "EXPIRED_AUTHENTICATION",
"message": "The OAuth token used to make this call expired 7 second(s) ago.",
"statusCode": 401,
"statusMessage": "Unauthorized"

Reason for Error: Because the token was expired on the first request, we refreshed the token and then made the request again, see below.

Response 2:

"message": "resource not found",
"statusCode": 404,
"statusMessage": "Not Found"

Reason for Error: On this request, the contact was not found, so we then proceeded with the second request to attempt to create the contact. 

Response 3:

"category": "CONFLICT",
"message": "Contact already exists. Existing ID: 11430351",
"statusCode": 409,
"statusMessage": "Conflict"

Reason for Error: This final request failed because of a Conflict error stating that the contact already exists. This is somewhat counterintuitive, being that when we first tried to update an existing contact, the response indicated that the contact did not exist. Read more from HubSpot [here](https://community.hubspot.com/t5/APIs-Integrations/exceptions-ConflictException-Contact-already-exists-Existing-ID/m-p/399522).


src/connections/destinations/catalog/actions-hubspot-cloud/index.md
```diff
@@ -67,6 +67,13 @@ Association Label | Select an association label between both the object types. F
 
 ## FAQ and troubleshooting
 
+### Why am I receiving a, "Contact already exists" error?
+This will only apply to integrations with two mappings that could create profiles in HubSpot. 
+1. Initially, the Upsert Contact action seeks to update an existing contact.
+2. If no contact is found, a subsequent attempt is made to create a new contact, potentially leading to three separate HubSpot API requests. For instance, an 'Expired Authentication' error may occur if the token expires on the initial request, prompting a token refresh and a subsequent request.
+3. If the next error indicates 'resource not found', it means the contact wasn't located, leading to a second attempt to create the contact. However, this attempt might fail due to a 'Conflict' error, suggesting the contact already exists. This situation can arise if another mapping is activated, causing the contact to be created by the time the Upsert Contact Action attempts its final contact creation request, due to the Custom Behavioral Event Action being triggered as well.
+Consequently, this error is displayed in the event delivery tab within Segment's UI.
```
stayseesong, line 75:
```suggestion
### Why am I receiving a `Contact already exists` error?
This error only applies to integrations with 2 mappings that can create profiles in HubSpot. Initially, the Upsert Contact action seeks to update an existing contact. If no contact is found, a subsequent attempt is made to create a new contact, potentially leading to 3 separate HubSpot API requests. For example, an `Expired Authentication` error may occur if the token expires on the initial request, prompting a token refresh and a subsequent request. If the next error indicates `resource not found`, it means the contact wasn't located, leading to a second attempt to create the contact. However, this attempt might fail due to a `Conflict` error, suggesting the contact already exists. This situation can arise if another mapping is activated, which causes the contact to be created by the time the Upsert Contact Action attempts its final contact creation request, due to the Custom Behavioral Event Action being triggered as well.
```

stayseesong (changes requested):

src/connections/destinations/catalog/actions-hubspot-cloud/index.md
```diff
@@ -65,7 +65,10 @@ Search Fields to associate |  This finds a unique record of custom object based
...
 
-## FAQ and troubleshooting
+## FAQs and troubleshooting
+
+### Why am I receiving a `Contact already exists` error?
+This error only applies to integrations with 2 mappings that can create profiles in HubSpot. Initially, the Upsert Contact action seeks to update an existing contact. If no contact is found, a subsequent attempt is made to create a new contact, potentially leading to 3 separate HubSpot API requests. For example, an `Expired Authentication` error may occur if the token expires on the initial request, prompting a token refresh and a subsequent request. If the next error indicates `resource not found`, it means the contact wasn't located, leading to a second attempt to create the contact. However, this attempt might fail due to a `Conflict` error, suggesting the contact already exists. This situation can arise if another mapping is activated, which causes the contact to be created by the time the Upsert Contact Action attempts its final contact creation request, due to the Custom Behavioral Event Action being triggered as well.
```
stayseesong, line 71:
```suggestion
This error only applies to integrations with 2 mappings that can create profiles in HubSpot. Initially, the Upsert Contact action seeks to update an existing contact. If no contact is found, a subsequent attempt is made to create a new contact, potentially leading to 3 separate HubSpot API requests. For example, an `Expired Authentication` error may occur if the token expires on the initial request, prompting a token refresh and a subsequent request. If the next error indicates `resource not found`, it means the contact wasn't located, leading to a second attempt to create the contact. However, this attempt might fail due to a `Conflict` error, suggesting the contact already exists. This situation can arise if you activate another mapping, which causes the contact to be created by the time the Upsert Contact Action attempts its final contact creation request, due to the Custom Behavioral Event Action being triggered as well.
```

stayseesong (approved):

stayseesong (approved):

PR #6248, opened 2024-03-19
Update GA4 Web docs

sarahrudy:
### Proposed changes
- add clarity around "Order Completed" event
- clarify that customers will need to add the "Purchase" action to their mappings in order for Segment to map "Order Completed" to GA4's "Purchase" event

### Merge timing
- ASAP once approved


cmastr:
closing as this has no changes now.

PR #6326, opened 2024-04-01
Niall/partner partial

niallzato:
### Proposed changes
Show partner status for partner owned integrations

### DIFF
--- a/scripts/catalog/updateDestinations.js
+++ b/scripts/catalog/updateDestinations.js
@@ -142,7 +142,8 @@ const updateDestinations = async () => {
         connection_modes,
         settings,
         actions,
-        presets
+        presets,
+        partnerOwned: destination.partnerOwned
       };
   
       // Add the updated destination to the destinationsUpdated array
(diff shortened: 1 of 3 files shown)


niallzato:
> hi @niallzato - I think this looks great! I did have one question: I built this locally to spot-check it, and I found that the new note only appears in some destinations. Is that intentional?

Correct, my thinking was to only show if for destinations that are partner owned, as the info is only really relevant for those destinations.
I thought seeing something like the following on destinations we own (GA4 etc) would be a bit redundant:
```
Partner Status
Not partner owned.
```

pwseg (approved):

forstisabella (approved):

forstisabella:
> > hi @niallzato - I think this looks great! I did have one question: I built this locally to spot-check it, and I found that the new note only appears in some destinations. Is that intentional?
> 
> @forstisabella correct, my thinking was to only show if for destinations that are partner owned, as the info is only really relevant for those destinations. I thought seeing something like the following on destinations we own (GA4 etc) would be a bit redundant:
> 
> ```
> Partner Status
> Not partner owned.
> ```

makes total sense - approved!

PR #6367, opened 2024-04-09
Update destination-filters.md | Correct execution order of Dest Filters

samkcrespo:
### Proposed changes

Correct execution order of Destination Filters

### DIFF
--- a/src/connections/destinations/destination-filters.md
+++ b/src/connections/destinations/destination-filters.md
@@ -23,7 +23,7 @@ Common use cases for destination filters include:
 Keep the following limitations in mind when you use destination filters:
 
 - Destination Filters aren't applied to events sent through the Event Tester.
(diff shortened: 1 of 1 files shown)


src/connections/destinations/destination-filters.md
```diff
@@ -23,7 +23,7 @@ Common use cases for destination filters include:
 Keep the following limitations in mind when you use destination filters:
 
 - Destination Filters aren't applied to events sent through the Event Tester.
-- Segment applies destination filters one at a time in the order that they appear in your workspace.
+- Segment applies destination filters in the following order: Sample, Drop ('Only Sends' are Drops), Drop Properties, Allow Properties
```
stayseesong, line 26:
```suggestion
- Segment applies destination filters in the following order: Sample, Drop ('Only Sends' are Drops), Drop Properties, Allow Properties.
```

stayseesong (approved):

PR #6464, opened 2024-04-30
Audience cycle error

Panandhan22:
### Why am I receiving the error "The audience would create a cycle by referencing another audience"?

This error occurs when creating audiences that reference each other, meaning audience X refers to audience Y in its trigger condition, and later you attempt to modify audience Y's trigger condition to refer back to audience X. To avoid this error, ensure that the audiences do not reference each other in their conditions.


Panandhan22:
Opened another pull request for this one.
https://github.com/segmentio/segment-docs/pull/6545

PR #6518, opened 2024-05-08
Update index.md

samkcrespo:
### Proposed changes

Included example code for verifying the x-signature of batch requests

### DIFF
--- a/src/connections/destinations/catalog/webhooks/index.md
+++ b/src/connections/destinations/catalog/webhooks/index.md
@@ -217,6 +217,24 @@ if (signature === digest) {
 }
 ```
 
+For Batch events, the process to authenticate these requests slightly differs as it involves verifying the X-Signature header against a hash of the **first event** in the batch.
+
+An example of how one might authenticate batch requests would be:
+
+```javascript
+ const signature = req.headers['x-signature'];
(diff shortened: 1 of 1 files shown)


src/connections/destinations/catalog/webhooks/index.md
```diff
@@ -217,6 +217,24 @@ if (signature === digest) {
 }
 ```
 
+For Batch events, the process to authenticate these requests slightly differs as it involves verifying the X-Signature header against a hash of the **first event** in the batch.
```
stayseesong, line 220:
```suggestion
For Batch events, the process to authenticate these requests slightly differs as it involves verifying the X-Signature header against a hash of the *first event* in the batch.
```

src/connections/destinations/catalog/webhooks/index.md
```diff
@@ -217,6 +217,24 @@ if (signature === digest) {
...
+
+An example of how one might authenticate batch requests would be:
```
stayseesong, line 222:
```suggestion
An example of how you might authenticate batch requests would be:
```

stayseesong (changes requested):

stayseesong (approved):

PR #6585, opened 2024-05-22
resizing image to reduce blur [netlify-build]

cmastr:
https://segment.atlassian.net/browse/DOC-918


src/unify/linked-profiles/setup-guides/snowflake-setup.md
```diff
@@ -20,7 +20,7 @@ Log in to Snowflake with a user that has admin privileges to provide Segment Lin
 
 Segment requires the following settings to connect to your Snowflake warehouse.
 
-![Connect Snowflake to the Data Graph](/docs/unify/images/snowflake-setup.png)
+<img src="/docs/unify/images/snowflake-setup.png" alt="AConnect Snowflake to the Data Graph" width="5888"/>
```
pwseg, line 23:
```suggestion
<img src="/docs/unify/images/snowflake-setup.png" alt="Connect Snowflake to the Data Graph" width="5888"/>
```

Did they ask us to use HTML tags instead of Markdown for images?

pwseg (approved):

src/unify/linked-profiles/setup-guides/snowflake-setup.md
cmastr: I did some research that suggested fixing the width of the image would help make it clearer (like setting the width to the same as what it was taken in). I tried that in markdown (see commit [#1](https://twilio.slack.com/archives/CL364MUJD)) and it didn't work so I used HTML instead.I did some research that suggested fixing the width of the image would help make it clearer (like setting the width to the same as what it was taken in). I tried that in markdown (see commit [#1](https://twilio.slack.com/archives/CL364MUJD)) and it didn't work so I used HTML instead.

PR #6616, opened 2024-05-29
Add faq for App Name

niallzato:
### Proposed changes
App Name is a required field, but not all our libraries add it to the payload. Calling this out in the FAQ section.

### DIFF
--- a/src/connections/destinations/catalog/actions-pinterest-conversions-api/index.md
+++ b/src/connections/destinations/catalog/actions-pinterest-conversions-api/index.md
@@ -74,9 +74,9 @@ Conversion Events must meet the following requirements to be considered for dedu
 > info ""
 > Segment offers a client-side destination specifically designed for the Pinterest Tag. You can find detailed documentation and further information on how to implement this integration by following this [link](https://segment.com/catalog/integrations/pinterest-tag/){:target="_blank”}.
 
-### Events failing to send due to no App Name set
+### Events fail to send due to no App Name set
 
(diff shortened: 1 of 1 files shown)


src/connections/destinations/catalog/actions-pinterest-conversions-api/index.md
```diff
@@ -74,6 +74,10 @@ Conversion Events must meet the following requirements to be considered for dedu
 > info ""
 > Segment offers a client-side destination specifically designed for the Pinterest Tag. You can find detailed documentation and further information on how to implement this integration by following this [link](https://segment.com/catalog/integrations/pinterest-tag/){:target="_blank”}.
 
+### Events failing to send due to no App Name set
```
forstisabella, line 77:
```suggestion
### Events fail to send due to no App Name set
```

src/connections/destinations/catalog/actions-pinterest-conversions-api/index.md
```diff
@@ -74,6 +74,10 @@ Conversion Events must meet the following requirements to be considered for dedu
...
+
+App Name is a mandatory field for many of the mappings. For Segment's mobile libraries, the App Name will be automatically collected and mapped to the correct field. However, for our web or server-based libraries, we do not automatically collect this field, which can cause the mapping to fail. We recommend adding the App Name to the Segment event, or hardcoding a static string in the mapping as the App Name.
```
forstisabella, line 79:
```suggestion
App Name is a mandatory field for many of the Pinterest Conversion API destination's mappings. Segment's mobile libraries automatically collect and map the App Name to the correct field. However, Segment's web or server-based libraries do not automatically collect this field, which can cause mappings to fail. Segment recommends adding the App Name to the Segment event, or hardcoding a static string in the mapping as the App Name.  
```

PR #6634, opened 2024-05-31
Add FAQ about destination filters

ashton-huxtable:
### Proposed changes

We occasionally get questions about why the event tester is not accounting for destination filters. The event tester bypasses these filters and is not designed to test for them. I added an FAQ to the docs to cover this topic. 

### Merge timing
- ASAP once approved

### Related issues (optional)

n/a


src/connections/test-connections.md
```diff
@@ -63,3 +63,7 @@ If you experience an error, [let Segment know](mailto:friends@segment.com) and t
 #### Is this feature available for Data Lakes?
 
 The Event Tester is not available for Data Lakes.
+
+#### Why are my destination filters being ignored?
+
+Events passed into the Event Tester bypass destination filters. Destination filters are applied to events as they are sent to specific destinations based on your configuration. However, the Event Tester is designed to help you troubleshoot your Sources, their configuration, and their downstream destinations by showing a sample of the data available. It allows you to check that data is being sent, and that it's in the correct format without the filters being applied. This means that when you use the Event Tester, you're seeing the data before any destination filters or other processing rules are applied, providing a clear view of the raw event data as it comes from the source.
```
forstisabella, line 69: "based on your configuration" - is this referring only to the destination filters, or does this refer to all tools that someone could enable to impact delivery (integrations object, consent management, etc.?)

ashton-huxtable: Destination filters would only apply to events that are passed to the destination, so this would be regardless of other tools that impact how data is sent downstream. Reviewing it, I can see that this line is a little confusing so we could remove the 'based on your configuration' rather than explaining further to keep it simple.

forstisabella, line 69:
```suggestion
Events passed into the Event Tester bypass destination filters. Destination filters are applied to events as they are sent to specific destinations. However, the Event Tester is designed to help you troubleshoot your Sources, their configuration, and their downstream destinations by showing a sample of the data available. It allows you to check that data is being sent, and that it's in the correct format without the filters being applied. This means that when you use the Event Tester, you're seeing the data before any destination filters or other processing rules are applied, providing a clear view of the raw event data as it comes from the source.
```

forstisabella (approved):

PR #6658, opened 2024-06-04
Update insert-functions.md Insert Functions only return one event

lizkane222:
### Proposed changes

It’s known to the support team that Insert Functions only return one event but it’s not clearly stated within Segment’s public docs.

### Merge timing
- ASAP once approved

### Related issues (optional)

Slack discussion : https://twilio.slack.com/archives/CA1AD07K5/p1717032800316919
Zendesk Ticket : https://segment.zendesk.com/agent/tickets/544419
KCS JIRA : https://segment.atlassian.net/browse/KCS-1442


src/connections/functions/insert-functions.md
```diff
@@ -183,6 +183,8 @@ You can read more about [error handling](#destination-insert-functions-logs-and-
...
... only return one event. This means that when the Insert Function receives an event then only one event will be sent to be handled by its configured mappings. However, if you would like [multiple mappings triggered by the same event](https://segment.com/docs/connections/destinations/actions/#:~:text=Multiple%20mappings%20triggered,Subscription%20Updated%20event.), _please see link's details_, you can do so by creating different types of mappings (Identify, Track, Page, etc) or multiple mappings of the same type, and then configuring the mapping's [trigger conditions](https://segment.com/docs/connections/destinations/actions/#conditions) to look for that event name/ type/ or other available field within the payload, and then configure the mapped fields to send different data. You can also configure the Insert Function to add additional data to be included within the event's payload before its handled by the mappings, and then configure the mapping's available fields to reference the payload's available fields. You may want to consider the event's [context object's](https://segment.com/docs/connections/spec/common/#context) available fields when adding new data to the event's payload.
```
forstisabella, line 186:
```suggestion
Unlike Source Functions and Destination Functions, which return multiple events, an Insert Function only returns one event. When the Insert Function receives an event, it sends the event to be handled by its configured mappings.

If you would like [multiple mappings triggered by the same event](/docs/connections/destinations/actions/#:~:text=Multiple%20mappings%20triggered,Subscription%20Updated%20event.):
1. Create different types of mappings (Identify, Track, Page, etc) or multiple mappings of the same type.
2. Configure the mapping's [trigger conditions](/docs/connections/destinations/actions/#conditions) to look for that event name/type or other available field within the payload.
3. Configure the mapped fields to send different data. 

You can also configure the Insert Function to add additional data to the event's payload before it's handled by the mappings and configure the mapping's available fields to reference the payload's available fields. 

You may want to consider the [context object's](/docs/connections/spec/common/#context) available fields when adding new data to the event's payload.
```

forstisabella (approved):

PR #6681, opened 2024-06-07
Freshsales Suite - traits is mandatory for Identify

joeynmq:
### Proposed changes
According to Freshsales's support team, traits are **mandatory** for Identify events:
> ​After conducting a more thorough investigation, we have discovered that the payload is missing the necessary traits, which is the root cause of the 400 error. It is important to note that traits are mandatory in the payload.

### Merge timing
ASAP once approved

### Related issues (optional)
https://segment.atlassian.net/browse/KCS-1451

### DIFF
--- a/src/connections/destinations/catalog/freshsales-suite-crm/index.md
+++ b/src/connections/destinations/catalog/freshsales-suite-crm/index.md
@@ -55,7 +55,7 @@ userId is a mandatory field which is used to identify the contact in Freshsales.
 
 #### Traits
 
-Traits are mandatory in an identify method, which contains pieces of information you know about a user.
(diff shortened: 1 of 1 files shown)


src/connections/destinations/catalog/freshsales-suite-crm/index.md
```diff
@@ -55,7 +55,7 @@ userId is a mandatory field which is used to identify the contact in Freshsales.
 
 #### Traits
 
-Traits are pieces of information you know about a user that are included in an identify method.
+Traits are mandatory in an identify method, which contains pieces of information you know about a user.
```
forstisabella, line 58:
```suggestion
Traits are pieces of information you know about a user. They are a mandatory part of the [Identify method](/docs/connections/spec/identify/).
```

forstisabella (approved):

PR #6726, opened 2024-06-19
Update scim.md

muhammadatifjav:
### Proposed changes

This change adds info about adding a flag to resolve a issue that has been reported by customers where a user removed from group will result in removing all users from a group.

Engineering team confirmed Microsoft Azure Entra ID (formerly Active Directory) is sending non-standard path value when issuing PATCH remove op. During testing, they tried out [the flag indicated in this Microsoft article](https://learn.microsoft.com/en-us/entra/identity/app-provisioning/application-provisioning-config-problem-scim-compatibility), which actually fixes the issue. It will be the fix to make Azure compatible with Segment's SCIM v2 implementation.

### Merge timing

- ASAP once approved

### Related issues 

https://segment.atlassian.net/browse/KCS-1426

### DIFF
--- a/src/segment-app/iam/scim.md
+++ b/src/segment-app/iam/scim.md
@@ -117,6 +117,9 @@ Instructions for configuring Microsoft Entra ID can be found on the Microsoft Do
 
 2. [Complete the Microsoft Entra ID setup guide for SCIM](https://learn.microsoft.com/en-us/entra/identity/saas-apps/segment-provisioning-tutorial){:target="_blank”}
 
+> info ""
+> To make Azure compatible with Segment's SCIM v2 implementation, append the flag `?aadOptscim062020` to the tenant URL as explained in [Microsoft Entra ID guide](https://learn.microsoft.com/en-us/entra/identity/app-provisioning/application-provisioning-config-problem-scim-compatibility#flags-to-alter-the-scim-behavior) . This change will ensure that the correct request structure is used when removing a user from a group.
+
 ## OneLogin Setup Guide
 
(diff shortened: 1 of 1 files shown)


src/segment-app/iam/scim.md
```diff
@@ -117,6 +117,9 @@ Instructions for configuring Microsoft Entra ID can be found on the Microsoft Do
 
 2. [Complete the Microsoft Entra ID setup guide for SCIM](https://learn.microsoft.com/en-us/entra/identity/saas-apps/segment-provisioning-tutorial){:target="_blank”}
 
+> info ""
+> To make Azure compatible with Segment's SCIM v2 implementation, append the flag `?aadOptscim062020` to the tenant URL as explained in [Microsoft Entra ID guide](https://learn.microsoft.com/en-us/entra/identity/app-provisioning/application-provisioning-config-problem-scim-compatibility#flags-to-alter-the-scim-behavior) . This change will ensure that the correct request structure is used when removing a user from a group.
```
forstisabella, line 121:
```suggestion
> To make Azure compatible with Segment's SCIM v2 implementation, append the flag `?aadOptscim062020` to the tenant URL as explained in the [Microsoft Entra ID documentation](https://learn.microsoft.com/en-us/entra/identity/app-provisioning/application-provisioning-config-problem-scim-compatibility#flags-to-alter-the-scim-behavior){:target="_blank”}. By appending the flag to your tenant URL, your request has the correct structure when you remove a user from a group.
```

forstisabella (approved):

PR #6738, opened 2024-06-24
Add private for shopify

tcgilbert:
### Proposed changes

add private so this can't be found

### Merge timing
asap


pwseg:
@tcgilbert So the `private` flag here doesn't necessarily hide the page (it looks like it's already hidden, actually).  If we add `private`, it means it's in private beta.  Is that the goal?

tcgilbert:
> @tcgilbert So the `private` flag here doesn't necessarily hide the page (it looks like it's already hidden, actually). If we add `private`, it means it's in private beta. Is that the goal?

Ah I see. Thanks for clarifying. I'll close this.

PR #6768, opened 2024-06-27
updated step to include a new 4th item

cmastr:
I wanted to add a 4th step to clarify this procedure for customers.


cmastr:
just doing a demo video.

PR #6789, opened 2024-07-03
Update Identity Resolution Onboarding

jamilmelhem:
Two critical changes:

1. Removed the info section with link the old Identity Resolution Settings pages - https://segment.com/docs/unify/identity-resolution/identity-resolution-settings/ 

We are using the "Onboarding" page for ID Res v2 and "Setting" page for ID Res v1. Customers not knowing which version they are reading and getting confused. Data is duplicate and having this at the top makes it even harder for customer to know which version they are reading. they are being mislead. Also, we do not need to publish ID Res v1 documentation any more; the settings "Setting" page should be deleted. there is only 1 production grade customers using ID Rev v1 but in any case v1 is forward compatible with v2. We already have a ticket to decom ID Res v1 

the section around segment automatically ordering Id res rules is not correct behavior ID Res v2. therefore, removing from documentation

### DIFF
--- a/src/unify/identity-resolution/identity-resolution-onboarding.md
+++ b/src/unify/identity-resolution/identity-resolution-onboarding.md
@@ -5,9 +5,6 @@ redirect_from:
   - "/personas/identity-resolution/identity-resolution-onboarding"
 ---
 
-> info ""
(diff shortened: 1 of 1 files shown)


pwseg (approved):

cmastr:
will update this topic with the rest of this JIRA: https://segment.atlassian.net/browse/DOC-962

PR #6840, opened 2024-07-22
Add info on troubleshooting Group properties not attribute Mixpanel groups

tanjinhong72:
### Proposed changes

Added the following to troubleshooting section.
-----------------------------------------------------------------
Group properties are not attributed to Mixpanel Groups

If the Mixpanel (Actions) destination uses $group_id as the group key, ensure that the mappings handling your `Group` events have the field for **Group ID** mapped to a valid value and the field for **Group Key** is left blank or without value. By default, the field **Group ID** is map to the event variable `groupId` and leaving the field for **Group Key** as blank will use $group_id as the group key.

### Related issues (optional)
https://segment.zendesk.com/agent/tickets/545764

### DIFF
--- a/src/connections/destinations/catalog/actions-mixpanel/index.md
+++ b/src/connections/destinations/catalog/actions-mixpanel/index.md
@@ -137,6 +137,10 @@ To send Track events with a custom Group Key, include the key as a property of T
 ```js
 analytics.track('Example Event', { custom_group_key : 'group1' });
 ```
+### Group properties are not attributed to Mixpanel Groups
+
+If the Mixpanel (Actions) destination uses $group_id as the group key, ensure that the mappings handling your `Group` events have the field for **Group ID** mapped to a valid value and the field for **Group Key** is left blank or without value. By default, the field **Group ID** is map to the event variable `groupId` and leaving the field for **Group Key** as blank will use $group_id as the group key. 
+
 ### Failed events due to timestamp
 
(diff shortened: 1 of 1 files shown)


src/connections/destinations/catalog/actions-mixpanel/index.md
```diff
@@ -137,6 +137,10 @@ To send Track events with a custom Group Key, include the key as a property of T
 ```js
 analytics.track('Example Event', { custom_group_key : 'group1' });
 ```
+### Group properties are not attributed to Mixpanel Groups
+
+If the Mixpanel (Actions) destination uses $group_id as the group key, ensure that the mappings handling your `Group` events have the field for **Group ID** mapped to a valid value and the field for **Group Key** is left blank or without value. By default, the field **Group ID** is map to the event variable `groupId` and leaving the field for **Group Key** as blank will use $group_id as the group key.
```
stayseesong, line 142:
```suggestion
If the Mixpanel (Actions) destination uses `$group_id` as the group key, ensure that the mappings handling your `Group` events have the **Group ID** mapped to a valid value the **Group Key** is left blank or without value. By default, the **Group ID** maps to the event variable `groupId`. If you leave the **Group Key** blank, the `$group_id` is used as the group key. 
```

stayseesong (changes requested):

PR #6848, opened 2024-07-23
Update Mutiny Destination to Remove support for page events

choncou:
### Proposed changes

Remove documentation about tracking "page" events because the Mutiny destination does not support these events

I'm not sure how to remove "Page" from the destination info card at the top of the page. It doesn't seem to be part of the markdown
CleanShot 2024-07-23 at 10 32 56@2x

### Merge timing
- ASAP once approved


pwseg (approved):

pwseg:
@choncou Thanks for opening this PR. To update the destination info card, we need to make a couple of back-end changes.  You should see those changes reflected on the Mutiny page by 2 August 2024 at the latest.

PR #6933, opened 2024-08-12
Replace generic link text with descriptive link text 🦆 

forstisabella:
### Proposed changes
Replaced all instances (except those in src/_data that were pulled from PAPI) of `[here]() `and `[click here]()` with text describing what the link points to, removed an integration that is out of business, some light Vale fixes

### Merge timing
after approval!


pwseg (approved):
I mean, this is outrageous.  Incredible work.

PR #6956, opened 2024-08-20
Data Graph GA Updates

pwseg:
### Proposed changes

- Fixed some links and converted the Data Graph page to the index file.

### Merge timing

- ASAP once approved.

PR #6973, opened 2024-08-22
Update Side Navigation for Linked Audiences GA

pwseg:
### Proposed changes

- Added Linked Audiences and Data Graph pages to the site side navigation.

### Merge timing

- ASAP once approved.

### DIFF
--- a/src/_data/sidenav/main.yml
+++ b/src/_data/sidenav/main.yml
@@ -345,7 +345,7 @@ sections:
     - path: /unify/data-graph/
       title: Data Graph
     - section_title: Setup Guides
-      slug: unify/data-graph/setup-guides/
+      slug: /unify/data-graph/setup-guides/
       section:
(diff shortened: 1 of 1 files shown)


src/_data/sidenav/main.yml
```diff
@@ -339,6 +339,18 @@ sections:
       title: Sample Queries
     - path: /unify/profiles-sync/tables
       title: Tables & Materialized Views
+  - section_title: Data Graph
+    slug: unify/data-graph
+    section:
+    - path: /unify/data-graph/
```
stayseesong, line 345: I think this can be deleted

stayseesong (approved):

src/_data/sidenav/main.yml
pwseg: This will work once the Data Graph page gets renamed, which will happen with #6956.

PR #6980, opened 2024-08-26
Added FAQ Section Including Common Question/Answer Regarding Timestamp Discrepancies

prigiattiperrut:
### Proposed changes
* Introduced a new FAQ section in the documentation.
* Included a detailed explanation of common causes for timestamp discrepancies when using Segment's client-side libraries.

### Merge timing
ASAP once approved

### Related issues (optional)

Slack: https://twilio.slack.com/archives/CALA7QMJQ/p1724679975057569

### DIFF
--- a/src/connections/spec/common.md
+++ b/src/connections/spec/common.md
@@ -298,3 +298,25 @@ Segment calculates `timestamp` as `timestamp = receivedAt - (sentAt - originalTi
 
 > info ""
 > For client-side tracking it's possible for the client to spoof the `originalTimeStamp`, which may result in a calculated `timestamp` value set in the future.
+>
+
+## FAQ
+
+### Why Are Events Received with Timestamps Set in the Past or Future?
+
+If you're using one of Segment's client-side libraries, please note that several factors can cause timestamp discrepancies in your event data.
+
+1. **Overriding Timestamp Value:**  
+   - When a manual timestamp is set in the payload with a date in the past, it can cause events to appear as if they were sent earlier than they actually were.
+
+2. **Analytics.js Source with Retries Enabled:**  
+   - The [Retries](https://segment.com/docs/connections/sources/catalog/libraries/website/javascript/#retries) feature supports offline traffic by queuing events in Analytics.js. These events are sent or retried later when an internet connection is available, keeping the original timestamp intact.
+
+3. **Mobile App Backgrounded or Closed:**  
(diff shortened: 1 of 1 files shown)


alanjcharles (approved):
amazing- thank you!!

silesky (approved):

src/connections/spec/common.md
```diff
@@ -298,3 +298,25 @@ Segment calculates `timestamp` as `timestamp = receivedAt - (sentAt - originalTi
...
+   - The [Retries](https://segment.com/docs/connections/sources/catalog/libraries/website/javascript/#retries) feature supports offline traffic by queuing events in Analytics.js. These events are sent or retried later when an internet connection is available, keeping the original timestamp intact.
+
+3. **Mobile App Backgrounded or Closed:**  
+   - If a user closes the app, events may be queued within the app. These queued events won't be sent until the app is re-opened, potentially in the future, leading to timestamp discrepancies.
+
+4. **Inaccurate Browser/Device Clock Settings:**  
+   - Timestamps can be incorrect if the client's device time is inaccurate, as the `originalTimestamp` relies on the client device's clock, which can be manually adjusted.
+
+5. **Traffic from Internet Bots:**  
+   - [Internet Bots](https://segment.com/docs/guides/ignore-bots/#whats-a-bot) can sometimes send requests with unusual timestamps, either intentionally or due to incorrect settings, leading to discrepancies.
```
stayseesong, line 322:
```suggestion
* **Overriding Timestamp Value:**  When a manual timestamp is set in the payload with a date in the past, it can cause events to appear as if they were sent earlier than they actually were.
* **Analytics.js Source with Retries Enabled:** The [Retries](/docs/connections/sources/catalog/libraries/website/javascript/#retries) feature supports offline traffic by queuing events in Analytics.js. These events are sent or retried later when an internet connection is available, keeping the original timestamp intact.
* **Mobile App Backgrounded or Closed:** If a user closes the app, events may be queued within the app. These queued events won't send until the app is re-opened, potentially in the future, leading to timestamp discrepancies.
* **Inaccurate Browser/Device Clock Settings:** Timestamps can be incorrect if the client's device time is inaccurate, as the `originalTimestamp` relies on the client device's clock, which can be manually adjusted.
* **Traffic from Internet Bots:**  [Internet Bots](/docs/guides/ignore-bots/#whats-a-bot) can sometimes send requests with unusual timestamps, either intentionally or due to incorrect settings, leading to discrepancies.
```

stayseesong (changes requested):

stayseesong (approved):

PR #6999, opened 2024-09-03
Update roles.md

pthorat-twilio:
New Information added to PII access:

For a user with PII Access and Source Admin/Read-Only permission, it means it there are PII's in the Sources/Debugger, they would be able to view it. However, this doesn't translate to Privacy Portal access.

The Privacy Portal is only accessible by Workspace owners.

### Proposed changes

The [PII access](https://segment.com/docs/segment-app/iam/roles/#pii-access) is a part of Privacy portal, however PII permissions doesn't translate to Privacy Portal access, and it wasn't clearly mentioned earlier.

### DIFF
--- a/src/segment-app/iam/roles.md
+++ b/src/segment-app/iam/roles.md
@@ -81,9 +81,9 @@ The Segment App doesn't show detected Personally Identifiable Information (PII) 
 
 Workspace Owners can grant specific individuals or groups access to PII from their Access Management settings. PII Access only applies to the resources a user or user group has access to; it doesn't expand a user's access beyond the original scope. All Workspace Owners have PII access by default.
 
-For a user with PII Access and Source Admin/Read-Only permission, it means there are PII's in the Sources/Debugger and they would be able to view it. However, this doesn't translate to Privacy Portal access.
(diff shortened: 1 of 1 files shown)


src/segment-app/iam/roles.md
```diff
@@ -81,6 +81,10 @@ The Segment App doesn't show detected Personally Identifiable Information (PII)
 
 Workspace Owners can grant specific individuals or groups access to PII from their Access Management settings. PII Access only applies to the resources a user or user group has access to; it doesn't expand a user's access beyond the original scope. All Workspace Owners have PII access by default.
 
+For a user with PII Access and Source Admin/Read-Only permission, it means there are PII's in the Sources/Debugger and they would be able to view it. However, this doesn't translate to Privacy Portal access.
+
+The Privacy Portal is only accessible by Workspace owners.
```
forstisabella, line 86:
```suggestion
Users with PII Access and Source Admin/Read-Only permission, it means there are PII's in the Sources/Debugger and they would be able to view it. However, this doesn't translate to Privacy Portal access.

Only users with the Workspace Owner role can access the Privacy Portal.
```

forstisabella (approved):

PR #7024, opened 2024-09-11
Adding a note about "initialized" event and track() callback 

prigiattiperrut:
### Proposed changes

1. Added clarification on detecting when Analytics.js has loaded:
- Instead of relying on the ready() method, users can listen for the initialize event (window.analytics.initialized). This event returns true even when a destination is blocked, providing a more reliable way to detect library loading.

2. Updated function behavior after a 300 ms timeout:
- Clarified that the function, which runs after a 300 ms delay to give the browser time to make outbound requests, may not execute if one of the device-mode libraries is blocked from loading.

### Merge timing
ASAP once approved

### DIFF
--- a/src/connections/sources/catalog/libraries/website/javascript/index.md
+++ b/src/connections/sources/catalog/libraries/website/javascript/index.md
@@ -216,7 +216,7 @@ The `page` call has the following fields:
 | `name`       | optional | String   | The name of the page.                                                                                                                                                                                                                                |
 | `properties` | optional | Object   | A dictionary of properties of the page. Note: Analytics.js collects `url`, `title`, `referrer` and `path` are automatically. This defaults to a `canonical url`, if available, and falls back to `document.location.href`.                           |
 | `options`    | optional | Object   | A dictionary of options. For example, [enable or disable specific destinations](#managing-data-flow-with-the-integrations-object) for the call. _Note: If you do not pass a `properties` object, pass an empty object (like '{}') before `options`_. |
-| `callback`   | optional | Function | A function that runs after a timeout of 300 ms, giving the browser time to make outbound requests first.                                                                                                                                             |
+| `callback`   | optional | Function | A function that runs after a timeout of 300 ms, giving the browser time to make outbound requests first. However, this function may not execute if one of the device-mode libraries has been blocked from loading.                                                                                                                                           |
 
 #### Default page properties
 
@@ -362,7 +362,7 @@ analytics.identify("hello world")
 
 The `ready` method lets you pass in a method that gets called after Analytics.js finishes initializing and after all enabled device-mode destinations load. It's like [jQuery's `ready` method](https://api.jquery.com/ready/){:target="_blank"}, except for Destinations. Because it doesn't fire until all enabled device-mode destinations are loaded, it can't be used to change configuration options for downstream SDKs. That can only be done if the SDK is loaded natively. 
 
-The `ready` method isn't invoked if any Destination throws an error (for example, for an expired API key, incorrect settings configuration, or when a Destination is blocked by the browser) during initialization.
+The `ready` method isn't invoked if any Destination throws an error (for example, for an expired API key, incorrect settings configuration, or when a Destination is blocked by the browser) during initialization. If you're looking to detect when Analytics.js has loaded, instead of using the `ready` method, you can listen for the `initialize` event to be emitted (`window.analytics.initialized`). This event will return true even when a destination is blocked.
 
 The code in the `ready` function only executes after `ready` is emitted. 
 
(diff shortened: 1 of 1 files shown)


src/connections/sources/catalog/libraries/website/javascript/index.md
```diff
@@ -216,7 +216,7 @@ The `page` call has the following fields:
...
 | `options`    | optional | Object   | A dictionary of options. For example, [enable or disable specific destinations](#managing-data-flow-with-the-integrations-object) for the call. _Note: If you do not pass a `properties` object, pass an empty object (like '{}') before `options`_. |
-| `callback`   | optional | Function | A function that runs after a timeout of 300 ms, giving the browser time to make outbound requests first.                                                                                                                                             |
+| `callback`   | optional | Function | A function that runs after a timeout of 300 ms, giving the browser time to make outbound requests first. However, this function may not execute if one of the device-mode libraries has been blocked from loading.                                                                                                                                           |
```
forstisabella, line 219:
```suggestion
| `callback`   | optional | Function | A function that runs after a timeout of 300 ms, giving the browser time to make outbound requests first. However, this function might not execute if one of the device-mode libraries has been blocked from loading.                                                                                                                                           |
```

src/connections/sources/catalog/libraries/website/javascript/index.md
```diff
@@ -362,7 +362,7 @@ analytics.identify("hello world")
 
 The `ready` method lets you pass in a method that gets called after Analytics.js finishes initializing and after all enabled device-mode destinations load. It's like [jQuery's `ready` method](https://api.jquery.com/ready/){:target="_blank"}, except for Destinations. Because it doesn't fire until all enabled device-mode destinations are loaded, it can't be used to change configuration options for downstream SDKs. That can only be done if the SDK is loaded natively. 
 
-The `ready` method isn't invoked if any Destination throws an error (for example, for an expired API key, incorrect settings configuration, or when a Destination is blocked by the browser) during initialization.
+The `ready` method isn't invoked if any Destination throws an error (for example, for an expired API key, incorrect settings configuration, or when a Destination is blocked by the browser) during initialization. If you're looking to detect when Analytics.js has loaded, instead of using the `ready` method, you can listen for the `initialize` event to be emitted (`window.analytics.initialized`). This event will return true even when a destination is blocked.
```
forstisabella, line 365:
```suggestion
The `ready` method isn't invoked if any Destination throws an error (for example, for an expired API key, incorrect settings configuration, or when a Destination is blocked by the browser) during initialization. If you're looking to detect when Analytics.js has loaded, instead of using the `ready` method, you can listen for the `initialize` event to be emitted (`window.analytics.initialized`). This event returns `true` even when a destination is blocked.
```

forstisabella (approved):

src/connections/sources/catalog/libraries/website/javascript/index.md
```diff
@@ -362,7 +362,7 @@ analytics.identify("hello world")
 
 The `ready` method lets you pass in a method that gets called after Analytics.js finishes initializing and after all enabled device-mode destinations load. It's like [jQuery's `ready` method](https://api.jquery.com/ready/){:target="_blank"}, except for Destinations. Because it doesn't fire until all enabled device-mode destinations are loaded, it can't be used to change configuration options for downstream SDKs. That can only be done if the SDK is loaded natively. 
 
-The `ready` method isn't invoked if any Destination throws an error (for example, for an expired API key, incorrect settings configuration, or when a Destination is blocked by the browser) during initialization.
+The `ready` method isn't invoked if any Destination throws an error (for example, for an expired API key, incorrect settings configuration, or when a Destination is blocked by the browser) during initialization. If you're looking to detect when Analytics.js has loaded, instead of using the `ready` method, you can listen for the `initialize` event to be emitted (`window.analytics.initialized`). This event returns `true` even when a destination is blocked.
```
JohnMealy23, line 365: I'm confused by this line.  Generally when the terms "listen", "emit" and "event" are used, it's in reference to a pubsub system.  Such a system would be ideal.  Something like...
```
window.analytics.on('initialized', function(event, properties, options) {
  console.log(event, properties, options)
});
```
However, I'm not seeing any way of subscribing here.  To date, `window.analytics.initialized` is a single boolean value.  The only way I can see to "listen" to it would be to set up an interval that poles for the value to change.  This is do-able (albeit inelegant, IMO).

If I'm correct, and there is no way of subscribing to changes in `window.analytics.initialized`, I'd strongly suggest updating this doc for clarity.  And if there _is_ a way of subscribing, I'd suggest linking to that documentation here.

prigiattiperrut: Hey @JohnMealy23, thanks for your note! I've created a new PR to address the adjustment: 
https://github.com/segmentio/segment-docs/pull/7036/files

PR #7076, opened 2024-09-25
Update snowflake-setup.md

charholt:
### Proposed changes

RETL docs, are missing allowlist info. Added the pre-built IP allowlist info.

### Merge timing
- ASAP once approved?


stayseesong:
@charholt is there a reason why we need this specifically on the Snowflake RETL setup page?

charholt:
It really should be on all the set-up pages except BQ, since allow listing is a requirement for almost all the connections.

stayseesong:
@charholt is this note still necessary to have in our docs? It's been almost a year. Did all users not update their allowlists?

stayseesong (approved):

PR #7114, opened 2024-10-04
Best Practices for Custom Domains

joeynmq:
### Proposed changes
We have been asked several times about setting up multiple custom domains for sources/environments/domains. So would like to add the best practices to the FAQ section for some clarifications.

### Merge timing
ASAP once approved

### Related issues (optional)
https://segment.atlassian.net/browse/KCS-1699

### DIFF
--- a/src/connections/sources/custom-domain.md
+++ b/src/connections/sources/custom-domain.md
@@ -59,6 +59,12 @@ To configure Custom Domain:
 
 ## FAQ
 
+### Can I set up multiple Custom Domains?
+For best practices, Segment recommends using only **one** custom domain per domain and only in **production** environments. It is not recommended to create multiple custom domains for the same source, and there are several reasons for this:
+
+- For default settings of the Analytics.js library, our customers only use a single segment.io - so having a single domain would be adequate.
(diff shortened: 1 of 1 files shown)


wally-tg (dismissed):

src/connections/sources/custom-domain.md
```diff
@@ -59,6 +59,9 @@ To configure Custom Domain:
 
 ## FAQ
 
+### Can I set up multiple Custom Domains?
+For best practices, Segment recommends using only **one** subdomain (e.g., `mysubdomain.mydomain.com`) per domain. Multiple custom domains for the same source are not supported.
```
forstisabella, line 63:
```suggestion
Segment recommends using only **one** subdomain (e.g., `mysubdomain.mydomain.com`) per source. Multiple custom domains for the same source are not supported.
```

forstisabella: @joeynmq - would this be per source or per workspace?

joeynmq: @forstisabella - this would be **per source** (more specifically, per domain, as we typically recommend setting up separate sources for different domains)

PR #7127, opened 2024-10-08
Unhide Use Cases Docs

pwseg:
### Proposed changes

- Unhid Use Cases docs.  They can be shared publicly now.
- Added Use Cases to the side navigation.
- Added a callout to a couple of the pages noting current Use Cases availability.

### Merge timing

- ASAP once approved.

### DIFF
--- a/src/getting-started/use-cases/guide.md
+++ b/src/getting-started/use-cases/guide.md
@@ -1,6 +1,5 @@
 ---
 title: Choosing a Use Case
-hidden: true
 ---
 
 Segment built Use Cases to streamline the process of implementing Segment for specific business objectives. 
--- a/src/getting-started/use-cases/guide.md
+++ b/src/getting-started/use-cases/guide.md
@@ -7,6 +7,9 @@ Segment built Use Cases to streamline the process of implementing Segment for sp
 
 This guide will help you navigate through the available use cases and select the one that best aligns with your business goals. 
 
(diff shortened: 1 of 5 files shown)


src/getting-started/use-cases/guide.md
```diff
@@ -1,12 +1,14 @@
...
 title: Choosing a Use Case
-hidden: true
 ---
 
 Segment built Use Cases to streamline the process of implementing Segment for specific business objectives. 
 
 This guide will help you navigate through the available use cases and select the one that best aligns with your business goals. 
 
+> info ""
+> You can onboard to Segment with a Use Case if you’re a new business tier user or haven’t yet connected a source and destination.
```
forstisabella, line 10: is Business Tier a proper noun? I looked for it on the P&P page but couldn't find it anywhere: https://segment.com/pricing/

```suggestion
> You can onboard to Segment with a Use Case if you’re a new Business Tier user or haven’t yet connected a source and destination.
```

forstisabella (dismissed):

src/getting-started/use-cases/guide.md
pwseg: @forstisabella No, that's a good catch, they should be capitalized.  Weirdly, we also describe it as "business plan" in other places.

forstisabella (approved):

PR #7144, opened 2024-10-15
array info for Triggers

spencerattick:
### Proposed changes

A common misconception is that Actions Triggers support filtering on arrays when they, unfortunately, don't. This update clarifies that.

### Merge timing
ASAP is fine


src/connections/destinations/actions.md
```diff
@@ -38,7 +38,8 @@ A Destination Action contains a hierarchy of components, that work together to e
...
-| Triggers        | Enable you to define *when* the corresponding Action fires. As part of a Trigger, you can use condition-based filters to narrow the scope of the Trigger. <br /> <br /> Self-service users can add a maximum of two conditions per Trigger.                                                                                                                                                             |
+| Triggers        | Enable you to define *when* the corresponding Action fires. As part of a Trigger, you can use condition-based filters to narrow the scope of the Trigger. <br /> <br /> Self-service users can add a maximum of two conditions per Trigger.  <br /> <br /> Triggers do not currently support matching on event fields containing ‘.$’ or ‘.$.’, which reference an array type.
```
stayseesong, line 41:
```suggestion
| Triggers        | Enable you to define *when* the corresponding Action fires. As part of a Trigger, you can use condition-based filters to narrow the scope of the Trigger. <br /> <br /> Self-service users can add a maximum of two conditions per Trigger.  Triggers don't support matching on event fields containing ‘.$’ or ‘.$.’, which reference an array type.
```

src/connections/destinations/actions.md
```diff
@@ -38,7 +38,8 @@ A Destination Action contains a hierarchy of components, that work together to e
...
-| Triggers        | Enable you to define *when* the corresponding Action fires. As part of a Trigger, you can use condition-based filters to narrow the scope of the Trigger. <br /> <br /> Self-service users can add a maximum of two conditions per Trigger.                                                                                                                                                             |
+| Triggers        | Enable you to define *when* the corresponding Action fires. As part of a Trigger, you can use condition-based filters to narrow the scope of the Trigger. <br /> <br /> Self-service users can add a maximum of two conditions per Trigger.  Triggers don't support matching on event fields containing ‘.$’ or ‘.$.’, which reference an array type.
+                                                                                                                                                             |
```
stayseesong, line 42:
```suggestion
```

PR #7173, opened 2024-10-28
BigQuery Data Graph Public Beta

stayseesong:


src/unify/data-graph/setup-guides/BigQuery-setup.md
```diff
@@ -1,54 +1,85 @@
...
+4. Enter your warehouse credentials. Segment requires the following settings to connect to your BigQuery warehouse:
+   * **Service Account Credentials:** JSON credentials for a GCP Service Account that has BigQuery read/write access. This is the credential created in Step 1.
+   * **Data Location:** This specifies the primary data location. This can be either region or multi-region.
+5. Test your connection, then click **Save**.
+
+## Update user access for Segment Reverse ETL dataset
+If you ran Segment Reverse ETL in the project you are configuring as the Segment connection project, a Segment-managed dataset is already created and you need to provide the new Segment user access to the existing dataset. 
+
+Grant the [BigQuery Data Editor](https://cloud.google.com/bigquery/docs/access-control#bigquery.dataEditor){:target="_blank"} role on the `__segment_reverse_etl` dataset to the service account if you run into an error on the Segment app indicating that the user doesn’t have sufficient privileges on an existing `__segment_reverse_etl` dataset. Note that the `__segment_reverse_etl` dataset is hidden in the console. Run the following SQL command:
```
pwseg, line 81:
```suggestion
Grant the [BigQuery Data Editor](https://cloud.google.com/bigquery/docs/access-control#bigquery.dataEditor){:target="_blank"} role on the `__segment_reverse_etl` dataset to the service account if you run into an error on the Segment app indicating that the user doesn’t have sufficient privileges on an existing `__segment_reverse_etl` dataset. Note that the `__segment_reverse_etl` dataset is hidden in the console. Run the following SQL command:  
```

Something about this paragraph confuses me.  So they only need to do this *if* they run into this error, right?  Could we maybe put "if you run into an error...." first?   I read this initially as... you have to do this.

pwseg (dismissed):
One minor suggestion, but LGTM!

src/unify/data-graph/setup-guides/BigQuery-setup.md
stayseesong, line 81:
```suggestion
If you run into an error on the Segment app indicating that the user doesn’t have sufficient privileges on an existing `__segment_reverse_etl` dataset, grant the [BigQuery Data Editor](https://cloud.google.com/bigquery/docs/access-control#bigquery.dataEditor){:target="_blank"} role on the `__segment_reverse_etl` dataset to the service account . Note that the `__segment_reverse_etl` dataset is hidden in the console. Run the following SQL command:  
```

pwseg (approved):

PR #7227, opened 2024-11-14
Update user-deletion-and-suppression.md

dineshraj-arunmozhi:
Adding a second option that the customer can use to update the data retention period in Segment workspace.

### Proposed changes

At the moment, in the public-facing documentation, we have just documented one option of updating the data retention period. It is a good idea to have both options options.


stayseesong:
@dineshraj-arunmozhi thank you for your suggestion, but we don't need to have 2 options.

PR #7274, opened 2024-11-29
Remove docs for deprecated Criteo Offline Conversions  integration

niallzato:
### Proposed changes
This integration is no longer supported as per our partner portal. Pr to simply remove our docs on it
image


forstisabella:
Hi @niallzato - when a destination is deprecated or otherwise unavailable, we typically just hide the docs (add `hidden: true` to the frontmatter). I can open a PR to do that, or if you want to open a PR I'd be happy to review it!

PR #7289, opened 2024-12-04
Update Use Cases Access Callouts [DOC-1017]

pwseg:
### Proposed changes

- Changed Use Cases access to callouts to "You can onboard to Segment with a Use Case if you’re a Business Tier customer." Use Cases are available to all Business Tier customers now.

### Merge timing

- ASAP once approved.


stayseesong (dismissed):
does "Use Case" need to be capitalized?

pwseg:
> does "Use Case" need to be capitalized?

Nah, you're right.  Just updated that.  Thanks for catching.

stayseesong (approved):

PR #7334, opened 2024-12-17
Update materialised tables

vanand17:
### Proposed changes

Currently, it is not clear what amount of historical data is available in materialised tables in our public docs

### Merge timing

- ASAP once approved?

### Related issues (optional)

NA

### DIFF
--- a/src/unify/profiles-sync/tables.md
+++ b/src/unify/profiles-sync/tables.md
@@ -266,6 +266,9 @@ With Profiles Sync, you can access the following three tables that Segment mater
 
 These materialized tables provide a snapshot of your Segment profiles, batch updated according to your sync schedule.
 
+> success ""
+> These tables will have two months of historical data on backfill.
+
 ### Switching to materialized Profile Sync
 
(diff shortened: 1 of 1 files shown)


src/unify/profiles-sync/tables.md
```diff
@@ -266,6 +266,9 @@ With Profiles Sync, you can access the following three tables that Segment mater
 
 These materialized tables provide a snapshot of your Segment profiles, batch updated according to your sync schedule.
 
+> success ""
+> Materialised view tables will have two months of historical data on backfill.
+
```
pwseg, line 271: @vanand17 Can you clarify what this means?

Does it mean that we only store two months of data at any time?  So if a user checks the table today, they'll only see data going back 60 days, and no further?

Or does it mean that the backfill process only retrieves two months of historical data when it **_first_** runs?  In other words, if someone sets up Profiles Sync today, the system would pull in the last 60 days of data, and from that point forward, it would keep accumulating new data.

PR #7359, opened 2024-12-27
Update test a destination action

vanand17:
### Proposed changes

Add details about why the exact event might not appear while testing a mapping using Step 2 in an actions destination 

### Merge timing

- ASAP once approved?

### Related issues (optional)

NA

### DIFF
--- a/src/connections/destinations/actions.md
+++ b/src/connections/destinations/actions.md
@@ -163,7 +163,11 @@ You can also test within the mapping itself. To test the mapping:
 1. Navigate to the **Mappings** tab of your destination. 
 2. Select a mapping and click the **...** and select **Edit Mapping**. 
 3. In step 2 of the mappings edit page, click **Load Test Event from Source** to add a test event from the source, or you can add your own sample event. 
-4. Scroll to step 4 on the page, and click **Test Mapping** to test the mapping and view the response from the destination. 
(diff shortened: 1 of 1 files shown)


src/connections/destinations/actions.md
```diff
@@ -163,7 +163,11 @@ You can also test within the mapping itself. To test the mapping:
 1. Navigate to the **Mappings** tab of your destination. 
 2. Select a mapping and click the **...** and select **Edit Mapping**. 
 3. In step 2 of the mappings edit page, click **Load Test Event from Source** to add a test event from the source, or you can add your own sample event. 
-4. Scroll to step 4 on the page, and click **Test Mapping** to test the mapping and view the response from the destination. 
+4. Scroll to step 4 on the page, and click **Test Mapping** to test the mapping and view the response from the destination.
+
+
+   >info ""
+   >The mechanism that allows that feature to pull an event from the pipeline only holds a small subset at a time and may not always find the event you're looking for. When that happens you can simply navigate to your source debugger, find an event to test with, copy that raw event, and paste the event into the Add test event interface.
```
forstisabella, line 170:
```suggestion

> info "Test Mapping might not return the events you're looking for"
> Segment only surfaces a small subset of events for the Test Mapping feature and might not always return the event you're looking for. If you'd like to test with a specific event, copy a specific event from your [Source Debugger](/docs/connections/sources/debugger/) and paste it into the **Add test event** interface.
```

PR #7386, opened 2025-01-15
Update portal.md

sneha-shashidhar:
### Proposed changes

Customer had reported that their workspace members without PII Access are able to view PII fields that are classified as Yellow. This was because there was no Custom Matcher created for those fields - so there was no logic for Segment to match and hash out data.

### DIFF
--- a/src/privacy/portal.md
+++ b/src/privacy/portal.md
@@ -227,7 +227,7 @@ Fields that are classified as 'Red' are masked for users that do not have PII Ac
 Keep in mind that if you have set Standard Controls to block fields from any of your sources, any new classifications you create in the Inbox will start to take affect immediately. For example, if you have a Privacy Control set up to block **Red** data from your Android source, any new fields you classify in the Inbox as **Red** will be blocked from entering Segment from your Android source.
 
 **Yellow Classification**:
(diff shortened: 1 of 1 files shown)


src/privacy/portal.md
```diff
@@ -227,7 +227,7 @@ Fields that are classified as 'Red' are masked for users that do not have PII Ac
 Keep in mind that if you have set Standard Controls to block fields from any of your sources, any new classifications you create in the Inbox will start to take affect immediately. For example, if you have a Privacy Control set up to block **Red** data from your Android source, any new fields you classify in the Inbox as **Red** will be blocked from entering Segment from your Android source.
 
 **Yellow Classification**:
-Fields that are classified as 'Yellow' are masked for users that do not have PII Access enabled. 
+Fields that are classified as 'Yellow' are masked for users that do not have PII Access enabled. A Custom Matcher is needed to mask fields other than those in the Default PII Matchers list.
```
stayseesong, line 230:
```suggestion
Fields that are classified as *Yellow* are masked for users that do not have PII Access enabled. You need a Custom Matcher to mask fields other than those in the Default PII Matchers list.
```

stayseesong (changes requested):

stayseesong (approved):

PR #7419, opened 2025-01-31
Added a clarification on Additional Domains

prigiattiperrut:
### Proposed changes
Clarified that multiple domains can exist within the same workspace while specifying that each source can only be associated with one domain, but a single domain can be linked to multiple sources.

### Merge timing
ASAP once approved

### DIFF
--- a/src/connections/sources/custom-domain.md
+++ b/src/connections/sources/custom-domain.md
@@ -50,7 +50,7 @@ To configure Custom Domain:
   - **Topic**: Select **Custom Domain**.
   - **Subject**: Enter a subject line for your support request.
   - **Domain Name**: Enter the subdomain that Segment should use for event request tracking.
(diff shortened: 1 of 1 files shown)


src/connections/sources/custom-domain.md
```diff
@@ -50,7 +50,7 @@ To configure Custom Domain:
   - **Topic**: Select **Custom Domain**.
   - **Subject**: Enter a subject line for your support request.
   - **Domain Name**: Enter the subdomain that Segment should use for event request tracking.
-  - **Additional Domain Name**: If applicable, add an additional subdomain. This field is optional.
+  - **Additional Domain Name**: If applicable, you can add an additional subdomain. This field is optional. You can have multiple domains within the same workspace; however, each source can only be associated with one domain. A single domain, however, can be associated with multiple sources.
```
stayseesong, line 53:
```suggestion
  - **Additional Domain Name**: (*Optional*) If applicable, you can add an additional subdomain. You can have multiple domains within the same workspace, however, each source can only be associated with one domain. A single domain can be associated with multiple sources.
```

stayseesong (changes requested):

stayseesong (approved):

PR #7438, opened 2025-02-13
Catalog Update 13 February 2025

pwseg:
Catalog update.


.bundle/config
```diff
@@ -1,2 +1,3 @@
 ---
 BUNDLE_PATH: "vendor/bundle"
+some change
```
forstisabella, line 3: ?

forstisabella (dismissed):

.bundle/config
pwseg: @forstisabella Omg, this was me tinkering.  Didn't realized I'd committed that, nice catch!

stayseesong (approved):

PR #7464, opened 2025-02-26
Data Graph docs fix traits to trait

stayseesong:


stayseesong:
@pwseg I think these were the only instances where `traits` was used instead of `trait`, but lmk if you see any other instances I missed

pwseg (approved):

PR #7673, opened 2025-06-11
Dot Notation Support in Mappings Destination Fields

wilwong-segment:
### Proposed changes
- Updated docs about new update with respect to dot notation support in destination field mappings and what is and isn't supported
- Rearranged the content a bit so that it's more clear

### Merge timing
Not to be released until we release this feature on June 12 2025


pooyaj (dismissed):
Perfect 👌

PR #7702, opened 2025-06-23
Add guidance on including/excluding Traits from Predictions

tbols17:
Updated wording to talk about ability to factor in traits into the models

### DIFF
--- a/src/unify/Traits/predictions/index.md
+++ b/src/unify/Traits/predictions/index.md
@@ -25,7 +25,7 @@ Follow these steps to build a prediction:
 - (For custom Predictive Goals) Add a condition(s) and event(s) to predict. 
 - Select the event and (optional) property that you want to use to make a prediction.
 5. Select a time period for the prediction. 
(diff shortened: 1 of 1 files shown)


src/unify/Traits/predictions/index.md
```diff
@@ -95,6 +95,13 @@ This table lists the requirements for a trait to compute successfully:
 
 Some customers want to specifically include or exclude events that get fed into the model. For example, if you track different events from an EU storefront compared to a US storefront and you only want to make predictions using data from the US, you could unselect the events from the EU space. This step is optional, Segment only recommends using it if you have a clear reason in mind for removing events from becoming a factor in the model.
 
+#### Selecting traits (optional)
```
pwseg, line 98: @tbols17 I rephrase some of this new section.  We usually avoid talking about what we "currently" support in the docs - customers usually think it's a guarantee that something will change in the future, when it may not.  Let me know if this looks good to you, and I'll get it merged.

pwseg (approved):

PR #7719, opened 2025-06-27
add info about custom traits/audience interaction

forstisabella:
### Proposed changes
Added a note about an edge case involving custom traits/consent that can cause some funky audience behavior

### Merge timing
asap!


rchinn1 (approved):
lgtm!

PR #7728, opened 2025-06-30
Decrease `font-weight` for bold text

forstisabella:
### Proposed changes
I got feedback from a PM that the current bold text "hurts the eyes. its jarring. some font styling has changed (may be in more than 1 spot) that makes reading hard and unpleasant" so I brought the font-weight value of bold text down one level: 

Current: 
Screenshot 2025-06-30 at 10 27 30 AM

Updated: 
Screenshot 2025-06-30 at 10 27 13 AM

### Merge timing
if approved


pwseg:
I like it!

PR #7754, opened 2025-07-09
remove mentions of Azure Data Lakes as Azure is going EOL

stayseesong:
### Merge timing
ASAP once approved


stayseesong:
> 🪦 RIP
> 
> could we replace instances of Segment Data Lakes (AWS) with just Segment Data Lakes, now that there's only one type?

good question! I'll check with the PM and see what they would like

PR #7832, opened 2025-08-13
Add Linked Audience Overview video

stayseesong:


pwseg (approved):

pwseg:
@stayseesong So they want to add it as is, no line introducing the video or anything?

stayseesong:
> @stayseesong So they want to add it as is, no line introducing the video or anything?

@pwseg yeah, the video doesn't need an introduction because the thumbnail of the video tells what it is, but if it's better for accessibility can add in a sentence

PR #8069, opened 2025-10-20
Add automated weekly report for new files 

pwseg:
### Proposed changes

- Added a GitHub Action that tracks new files added to the repo since mid-September.
- The platform team working on the migration needs a running list of new files since their initial snapshot.
- The action runs every Friday at noon ET and creates a GitHub issue with a list of all new files since September 17th.
- Each issue includes the file path/date added/commit link.
- I also created a migration label to add to this issue.
- We can also trigger the action manually from the Actions tab if we need an ad-hoc report.

### Merge timing

- ASAP once approved

### DIFF
--- a/.github/workflows/new-files-report.yml
+++ b/.github/workflows/new-files-report.yml
@@ -19,10 +19,10 @@ jobs:
       - name: Generate report
         id: report
         run: |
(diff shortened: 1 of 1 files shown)


forstisabella (approved):
baller

PR #8158, opened 2025-11-13
Updated Auto Retry Section.md

chiragtwilio:
Updated auto retry info box that mentioned that Auto-Retry may not be available in your workspace and it'll be available to all workspaces by 31 Jan, 2025. 

We've now released auto retry feature to all customers across tiers

Removed the callout on Auto Retry not being available for all workspaces as it has now been rolled out across customer tiers. 

### Merge timing
- ASAP once approved


stayseesong:
@chiragtwilio we've migrated over to using the Twilio Github, so I've transferred this PR over there

